PR #8, opened 2021-01-06
Performance Improvements 

samncorn:
Slightly more efficient use of Pandas functionality to conserve memory.


mschwamb:
@eggls6  Can you take a look. I think the conflicts can be merged, but probably best if you take a look

Issue #77, opened 2022-03-03
Calculating apparent magnitude colour correction for only the filter the observation is in


mschwamb:
Currently surveysimPP calculates the apparent magnitude in all filters and then later selects the specific filter. This should be modified to calculate only for the specific filter

fedorets:
Addressed by pull request #98.

PR #98, opened 2022-03-09
Update 09.03.2022 / GF

fedorets:
The code now calculates the colour offset for a single appropriate filter instead of all.

Resolves issue #77.


surveySimPP/modules/PPResolveMagnitudeInFilter.py
```diff
@@ -0,0 +1,49 @@
...
+    Output: updated padain
+    
+    Usage: padaout=PPResolveMagnitudeInFilter(padain,mainfilter,othercolours,resfilters)
+    
+    """
+    
+    apparent_mag=np.zeros(len(padain), dtype=float)
+    
+    i=0
+    while(i<len(resfilters)):
```
mschwamb, line 34: Please make this a for loop. It will be clearer in the future for someone trying to modify this in the future for i in np.arange (len(resfilters)):

surveySimPP/modules/PPResolveMagnitudeInFilter.py
```diff
@@ -0,0 +1,49 @@
...
+        inRelevantFilterList=(padain['optFilter']==resfilters[i])
+        inRelevantFilter=padain[inRelevantFilterList]
+        if(len(inRelevantFilter)>0):
+            if(resfilters[i]==mainfilter):
+               apparent_mag[inRelevantFilterList]=0.0
+            else:
+               apparent_mag[inRelevantFilterList]=inRelevantFilter[othercolours[i-1]]
```
mschwamb, line 41: I don't see how this step works correctly. If i starts out as 0 then in the first time through this you get for the index [i-1]=[0-1]=-1 which means you're polling the last item in the othercolors array when you want 0.  I think this should be i not i-1

mschwamb: Can you please check this is working correctly.

surveySimPP/modules/PPResolveMagnitudeInFilter.py
```diff
@@ -0,0 +1,49 @@
...
+        i=i+1
+    padain['apparent_mag']=apparent_mag
+    padain['MagnitudeInFilter'] = padain[mainfilter] + padain['apparent_mag']
```
mschwamb, line 44: there is no need to add another column for the pandas data frame and then delete it. 

`import pandas as pd`
`import numpy as np`
`df = pd.DataFrame({'points': [25, 12, 15, 14, 19, 23, 25, 29],
                   'assists': [5, 7, 7, 9, 12, 9, 9, 4],
                   'rebounds': [11, 8, 10, 6, 6, 5, 9, 12]})`

df['points']=df['points']+np.ones(len(df))

updates points by 1 as np.ones makes an array of ones the size of len(df). So this should just be padain['MagnitudeInFilter'] = padain[mainfilter] + apparent_mag

mschwamb: This is mixing pandas and numpy data frames so maybe bad things can happen with this so I am asking on the slack channel to those who are much more organized with pandas

mschwamb:
@fedorets some comments for you take a look at and address as needed

surveySimPP/modules/PPResolveMagnitudeInFilter.py
mschwamb: Siegfried says this is okay. So I think should change should be made so we're not adding structures to the pandas data frame that we're just deleting

surveySimPP/modules/PPResolveMagnitudeInFilter.py
fedorets: Ok.

surveySimPP/modules/PPResolveMagnitudeInFilter.py
fedorets: So, the way this works, is that in the config file we have three different variables:

1. mainfilter, to which everything else is compared
2. othercolours, which are the colour offsets, such as i-r
3. resfilters, which are ther results of colour manipulations. mainfilter is always the first one, with the other resulting filters following in the same order as othercolours. Therefore, len(resfilters) == len(othercolours) +1

The thoughts behind this admitedly somewhat clumsy structures are
1. there needs to be a place in the config file where we can see at a glance which all filters are included
2. there needs to be some way of matching colour offsets and resulting colours.

So for that structure everything should be correct. One thing how this situation can be omitted is pre-calculating all colours in the input parameter file. So instead of g-r, i-r etc we would have precalculated values for g, r, etc.

surveySimPP/modules/PPResolveMagnitudeInFilter.py
fedorets: Ok, makes sense!

surveySimPP/modules/PPResolveMagnitudeInFilter.py
mschwamb: That's dangerous to have resfilters and othercolours have different lengths. It makes it very easy for someone to forget this and do bad things with loops. othercolours[0] =0 for the main filter and the rest of the color terms following in the array would actually significantly simplify the code. You wouldn't need this bit  

`  if(resfilters[i]==mainfilter):
               apparent_mag[inRelevantFilterList]=0.0`

surveySimPP/modules/PPCalculateApparentMagnitude.py
```diff
@@ -18,30 +19,18 @@ def PPCalculateApparentMagnitude(observations, phasefunction, mainfilter, otherc
...
-         observations=PPhookBrightnessWithColour.PPhookBrightnessWithColour(observations, mainfilter, othercolours[i], resfilters[i+1])         
-         i=i+1
- 
-    observations=observations.reset_index(drop=True)
-
-    
-    pplogger.info('Resolving the apparent brightness in a given optical filter corresponding to the pointing...')
-    observations=PPMatchPointingsAndColours.PPMatchPointingsAndColours(observations,filterpointing)
+    pplogger.info('Selecting and applying correct colour offset...')
+    observations=PPResolveMagnitudeInFilter.PPResolveMagnitudeInFilter(observations,mainfilter,othercolours,resfilters)
```
mschwamb, line 34: This function needs a better name. 

How in reading the names of the functions would someone know the difference between PPCalculateApparentMagnitudeInFilter and PPResolveMagnitudeInFilter? They sound like exactly the same thing and they are not.

surveySimPP/modules/PPResolveMagnitudeInFilter.py
mschwamb: If you want to keep resfilter as with no entry for r, then this for loop needs to loop over othercolours and not resfilters. and pull the bit for the main filter out of the loop and do the main filter bit before the loop.  As written you're checking the logic statement each time of whether the main filter is the current filter being looked at in the loop. Whereas if you loop over othercolours and have resfilters[i+1] to access the correct array index then you will speed up the calculation because the logic statement for the main filter isn not being accessed every time.  We need every little speed up we can get

fedorets: How about I combine mainfilter and othercolours into a single array called inputcolours (also in the Config file)? This will force the colours that the user is using in one place, remove the possibility for errors, and get rid of the pesky if-statements?

So, we would have inputcolours=r,g-r,i-r and resfilters r,g,i.

fedorets: Changed per the second suggestion and added to the pull request.

Issue #110, opened 2022-03-19
modify outputformat config file parameters to take an separatelyCSV option


mschwamb:
Currently in the config file we have:

# options [csv | sqlite3 | hdf5 ]
outputformat = csv

# write out CSV file each object separately
separatelyCSV = False 

These can be combined into one where outputfot would be csv | sqlite3 | hdf5|separatelycsv and separatelycsv would serve the same purpose as separatelyCSV = True and would reduce confusion since output format would be on variable.

- [ ] Remove separatelyCSV  variable from config file read in 
- [ ] Add option for separatelyCSV  in outputformat 
- [ ]  make change to config file generator code

fedorets:
Fixed by pull request #129.

mschwamb:
Pull request not merged

PR #129, opened 2022-03-25
Update 25.03.2022 / GF / 2

fedorets:
Fixes issue #110 by making separatelyCSV an option for outputformat instead of its own parameter.


mschwamb:
There are lots of changes include lsstcomet/util.py → surveySimPP/lsstcomet/util.py

This pull request needs to be changed so that only the files impacted for these issues are updated not any others. Currently the pull request says 25 files were changed. That definitely is not the case. Please pull a fresh copy for the repo and modify again in a clean pull request and only upload to github the files that actual need changing

PR #166, opened 2022-04-06
Update 06.04.2022 / GF

### LINKED ISSUE #161: Update error handling for PPReadEphemerides.py
in https://github.com/dirac-institute/survey_simulator_post_processing/blob/8f5b34446c151c270e0ca92e308fe1a8fdfe6d94/surveySimPP/modules/PPReadEphemerides.py#L43
```
if (ephemerides_type=='oif'):
     padafr=PPReadOif.PPReadOif(eph_output, inputformat)
```

This should allow a capital OIF or a lower case oif. If there is any other entry for ephemerides_type other OIF or oif the code should stop. There should be an else after this if with an error reported if ephemerides_type was not set to a good value. Otherwise the error check letter will catch it but it won't be clear to the user that it is their config file that is wrong

fedorets:
Fixes #161 by enabling case-insensitivity and error handling.

- [ x] Does pip install still work?
- [x ] Does SurveySimPP run successfully on a test set of input files/databases?

### DIFF
--- a/demo/PPConfig.ini
+++ b/demo/PPConfig.ini
@@ -18,7 +18,7 @@ objecttype = asteroid
 # -b or --brightness or --phase     Brightness and phase parameter file name
  
 # type of input ephemerides: default = oif, options: oif
-ephemerides_type=ogre
+ephemerides_type=oif
 
 # pointing database
 pointingdatabase= ./data/test/baseline_10yrs_10klines.db 


mschwamb:
Dave found that we have circular imports of the logging, so he made changes to this .py. Can you reclone and make the changes to this version in the repo that has no circular import of the logging ?

PR #188, opened 2022-04-25
Standardising random number generation.

### LINKED ISSUE #154: change numpy random.random use numpy  default_rng that was setup in the main to other random calls
See https://numpy.org/doc/stable/reference/random/index.html#random-quick-start

astronomerritt:
Fixes #154 .

All random numbers used in the code are now generated using the random number generator defined in surveySimPP.py using the [recommended call ](https://numpy.org/doc/stable/reference/random/index.html#random-quick-start). This RNG is now passed to all functions which use random numbers.

These functions also have the option to use a default random number generator seeded with the system time if the RNG isn't passed to them.

I also apparently accidentally deleted the part of surveySimPP.py that performs the SSP criterion filtering? It's back in there now.


surveySimPP/modules/PPDetectionEfficiency.py
```diff
@@ -4,15 +4,19 @@
 import sys
 import logging
 import numpy as np
+import time
 
 # Author: Grigori Fedorets
 
-def PPDetectionEfficiency(padain, threshold):
+default_rng = np.random.default_rng(int(time.time()))
```
mschwamb, line 11: Why are there these call outside the function? This is now a global variable I believe which is being set multiple times.

mschwamb: We have a seed already set in surveySImPP so you should be able to send default_rng through, so why are these calls needed?

mschwamb (changes requested):
I don't think the seed needs to be set everything time as a global variable outside of the function, we set this in the main surveysimPP. In the tests, we're going to want to use a fixed seed to get a known answer.

surveySimPP/modules/PPDetectionEfficiency.py
astronomerritt: They're only a failsafe in case the functions are called without the RNG as an argument: this is how it's done already in PPRandomizeMeasurements.py so I was copying that for standardisation. I will happily remove these calls if you prefer.

mschwamb: yeah, that's making it a a global variable which is bad... means you can change it anywhere and can be called anywhere

https://www.programiz.com/python-programming/global-local-nonlocal-variables

if the function is called without the RNG argument it should crash because the arguments sent to the function are incorrect (see https://stackoverflow.com/questions/38364162/in-python-how-do-you-denote-required-parameters-and-optional-parameters-in-code for an explanation)

astronomerritt: Okay, no problem. I'll also remove the global from PPRandomizeMeasurements.py then.

mschwamb: Thanks. Otherwise I think every time the code is imported a new seed is generated, which we don't want. 

No global variables... (no wire hangers...)

astronomerritt:
Removed global variables and also made changes to surveySimPP.py to make it more PEP8 compliant via flake8. Several lines are still over 80 characters: these are the ones I couldn't find a good or elegant way to change.

PR #219, opened 2022-05-26
Clearing large number of issues.

### LINKED ISSUE #159: move output directory and outfilestem to command line and output log and error file to outpath
move 

# path to output
outpath = ./data/out/

# output file name stem
outfilestem = hundredcomets

from config file to command line so log is associated with the output directory and output log and error to outpath

### LINKED ISSUE #170: Tidy up PPConfigFileParser
PPRunUtilities.PPConfigFileParser has gotten very long and unwieldy. I'm hoping to eventually tidy up a lot of it into its own individual functions for readability and possibly pop it into its own module. Not a high priority as yet.

- [ ] add proper type checking
- [ ] general tidy up

### LINKED ISSUE #172: Check that all values being read from config file are the right type
Right now, if someone (for example) input a string (or somehting else that does not convert nicely to a float) for "brightLimit", it does not fail gracefully. I need to implement proper type checking.

### LINKED ISSUE #189: update random number generator for PPSimpleSensorArea.py
Have PPSimpleSensorArea use a predefined rng instance.

### LINKED ISSUE #190: call PPSimpleSensorArea from  PPFootprintFilter
If the fading function is not being applied and circle footprint and fill factor is less than 1 , then call [PPSimpleSensorArea.py](https://github.com/dirac-institute/survey_simulator_post_processing/blob/master/surveySimPP/modules/PPSimpleSensorArea.py)

### LINKED ISSUE #191: have flags for every switch for filter - don't use configs['columnname'] if it stores a value
if there is no bright limit in the config file then do not run the brightlimit  cut. 

I am concerned about statements like this:

```      
  if configs['brightLimit']:`
            observations = PPBrightLimit.PPBrightLimit(
                observations,
                configs['brightLimit'])
```

In some places we have boolean values being evaluated as flags like trailing losses and then like the above that aren't. Let's make the config have a flag for there to be a boolean true false set, because it's confusing that the value is changing between types if it exists or not. I know this is extra variables, but then everyone knows where we stand with that variable. Right now we're switching between types in surveySImPP. 

so have a brightlimit as the variable and a brightlimit run flag. Same thing for SNRLimit and maglimit. Because you evaluate  config_dict['SSPFiltering']  to be true or false and in the other if statements you are confirming there is a value in the variable which makes it true in SurveySimPP. This way all the if statements checking if  a filter should run are actually looking at booleans

### LINKED ISSUE #192: Put SurveySimPP function calls in new order
[Post_processing_plan_2.pdf](https://github.com/dirac-institute/survey_simulator_post_processing/files/8566871/Post_processing_plan_2.pdf)

Reorder SurveySimPP into the order of the flow chart to run the specific filters

### LINKED ISSUE #193: update config file parser functions to check if the variable is there without error trapping
```

def PPGetOrExit(config, section, key, message):
    # from Shantanu Naidu, objectInField
    try:
        return config[section][key]
    except KeyError:
        logging.error(message)
        sys.exit(message)
        

def PPGetOrPass(config, section, key, message):
    # as PPGetOrExit, except if the config variable doesn't exist, returns 0
    # essentially turns off filters if the variable isn't given
    try:
        return config[section][key]
    except KeyError:
        logging.info(message)
        return 0
```

There is config.has_option() and config.get() that can do these checks and set default values:

see https://stackoverflow.com/questions/70845111/configparser-check-if-variable-in-config-file-exists

Let's update these functions to do this and set a value. My concern is that if you 0 what happens if I set the value to 0 for one of these config file variables. Would we know the difference based on what's written. Using has_option() means we put flags true if the value or false if the value is there or not.

### LINKED ISSUE #195: Change SSP references to linking
Change SSP references to linking

astronomerritt:
Several changes made to the code in anticipation of the LSST sprint.

Fixes #159.
Output file location and filestem have been moved to the command line and logs are now saved in the output location.

Fixes #170. Fixes #172. 
Config file parser now tidier and checks to make sure all inputs are the correct type. 

Fixes #191. Fixes #193.
The configs dictionary now includes explicit True/False flags for all filters and checks for the presence of values using configparser's inbuilt functions.

Fixes #190. Fixes #189.
If the fading function is not being applied and the footprint is circle, PPApplyFOVFilter now calls PPSimpleSensorArea, which has been moved to PPApplyFOVFilter.py and now uses the random number generator object.

Fixes #192.
The functions in surveySimPP.py have been re-ordered to match the flow chart.

Fixes #195.
Renamed PPFilterSSPCriterionEfficiency to PPFilterSSPLinking, for clarity.

Additional fixes:
Fixed a typo in PPAddUncertainties.py on line 199 and cleaned up imports in surveySimPP.py. Made all files I edited PEP8 compliant.

- [x] Have you used flake8 on the files you have updated to confirm python programming style guide enforcement? (You can ignore line length warnings: flake8 --ignore=E501,E126,E228,E228,E133)


astronomerritt:
All files I edited are now PEP8 compliant according to flake8.

PR #230, opened 2022-06-06
Add testcase for ObjectsInField

fedorets:
Added instructions and a config file to be able to produce the ephemerides input file through objectsInField.

- [ ] Have you used flake8 on the files you have updated to confirm python programming style guide enforcement? (You can ignore line length warnings: flake8 --ignore=E501,E126,E228,E228,E133,W503)


mschwamb (changes requested):
No shell scripts. People should run the command on their own. Please revise

Why is awk being called? That's not then output that people would get if they just run OIF, let's not going changing things from what OIF would produce

PR #256, opened 2022-06-13
Improve packaging and configure tox

ConorMacBride:
This PR makes some improvements to the packaging of `surveySimPP`. I have moved from `setup.py` to `setup.cfg`. I have added `setuptools_scm` so it can automatically determine the version number from the latest git tag and also know which files to include when packaging. I've also added an initial tox configuration. We can use tox with GitHub Actions, but that can be a separate PR. Some more items:

- [Move test data into surveySimPP package](https://github.com/dirac-institute/survey_simulator_post_processing/commit/dcaea93c37b3ce180dcbfffd775b27903110aa6f)
  This way the test files are available within the package so the tests can be run in an isolated environment (for tox).
- [Add astropy helper function to get test filepath](https://github.com/dirac-institute/survey_simulator_post_processing/commit/1c3fe1093bb01786339d381361ec5a6033b9c5ab)
  The test files were hardcoded which meant that you had to run pytest from the root directory of the package. With the help of astropy, we can now look up the installed path to the data files during tests using a helper function.
- [Move utilities into surveySimPP](https://github.com/dirac-institute/survey_simulator_post_processing/commit/fc121645cc0ddb7863540dabb0a27533a85db242)
  This way you can do `import surveySimPP.utilities` instead of `import utilities` which has the possibility of conflicting with other packages.


ConorMacBride:
The reason there are so many tests failing is because I configured pytest (in `setup.cfg`) to also fail the tests if there are warnings. This way you can be made aware about things such as deprecation warnings and fix them before they become a real issue for your users. You can still choose warnings to ignore, like astropy do for example: https://github.com/astropy/astropy/blob/main/setup.cfg#L133-L158.

If #257 removes the warning, I'll rebase this PR to confirm all the tests pass before merging.

mschwamb:
Thanks @ConorMacBride. That's really helpful to catch the warnings.

PR #268, opened 2022-06-17
Fixing #65, #74, #223.

### LINKED ISSUE #65: Filter specified in config file but does not exist in colours file: uncaught error
If a filter is specified in the config file but does not exist in the colours file, an uncaught KeyError occurs in PPhookBrightnessWithColour, line 37. There needs to be a check done fairly early on to make sure that filters specified in the config file exist in the colours file.

### LINKED ISSUE #74: Verbose flag
Add command-line flag and if statements for verbosity. 

At the moment the default should probably be verbose==True but we can flip this after testing.

Question: should verbose==False only suppress console output but maintain a log? Or should it also suppress logging for everything but errors/warnings?

### LINKED ISSUE #223: Parametrise the width parameter for the fading function
The width parameter for the fading function should be parametrised in the PPConfig file generator.

Following the notation of Veres and Chesley (2017):

$$\epsilon(m) = \frac{F}{1 + e^\frac{m-m_5}{w}}$$

where epsilon is the probability of detection, F is the fill factor, m and m5 are observed and 5sigma magnitudes, and w is the width of the fading function. It turned out that there are two different values of the width parameter used in literature: Veres and Chesley use 0.1 and Jones et al. (2018) use 0.12, following SDSS. Post-processing is currently using 0.1.

Currently, the chain to the width parameter is:

surveySimPP.py:140:        observations = PPFilterFadingFunction(observations, configs['fillfactor'], rng)
modules/PPFilterFadingFunction.py:27:    observations["detection_probability"] = PPDetectionProbability(observations, fillFactor=fillfactor)

modules/PPDetectionProbability.py:45:def calcDetectionProbability(mag, limmag, fillFactor=1.0, w=0.1):
calcDetectionProbability(oif_df[magnitude_name],oif_df[limiting_magnitude_name], fillFactor, w)
modules/PPDetectionProbability.py:94:            return calcDetectionProbability(oif_df[magnitude_name] + oif_df[trailing_loss_name],oif_df[limiting_magnitude_name], fillFactor, w)

So this would require:
1. adding a new parameter called 'fadingFunctionWidth' to the config file generator.
2. Make sure in the PPUtilities it is understood when read in.
3. Pass it as a parameter to PPFilterFadingFunction in surveySimPP.
4. Pass it as a parameter to  calcDetectionProbability in PPDetectionProbability.

astronomerritt:
Fixes #65 .
PPReadPhysicalParameters now checks to make sure that the physical parameters file contains the colour offset columns which match the observing filters specified in the config file. If it doesn't, it errors out gracefully. Also changed many unit tests to reflect the fact that PPReadPhysicalParameters now takes 'othercolours' as an argument. All files I worked on are now PEP8 compliant.

Fixes #74.
There is now a verbose flag at the command line (-v). By default, it's True. If turned off, the only things printed to the log file are the command line arguments and config file variables for the run, in lieu of a header in the output.]

Fixes #223.
The width parameter for the fading function is now parametrised in the config file and passed to the relevant functions. The default is 0.1. If the fading function is turned off but the width parameter is still present, the code errors out gracefully. The value of the width parameter is error-handled to be greater than zero but less than 0.5.

I've also deleted PPRunUtilities and split what was in there into other files. Most of what was in there is now in PPConfigParser. Everything else got its own file. PPRunUtilities was too long and difficult to maintain. PPConfigParser is still too long, but it's less of a headache.

I've also moved all of the surveySimPP.modules imports in surveySimPP.py to their own file, surveySimPP_imports.py: it's tidier.

Also did some general PEP8ing and clarifying. Apologies for multiple open/closed pull requests, I was messing around with my branches.


mschwamb:
Let's not have a file with just imports. It makes it harder for someone reading the code because they have to refer to another file to see the imports.  I have concerns about making things tidier. Code files are allowed to be long.  and it's only 17 lines which you can easily move through. So I would prefer to have some more discussion of this and look at other software packages before making that change. Can you save that change somewhere else and adjust this pull request to not have that change.

mschwamb:
Thanks. Sorry if this felt urgent. Enjoy your weekend

astronomerritt:
No worries, it was only going to take me five minutes and I was on my laptop anyway. Happy to revert any changes you're not keen on in general :)

Issue #353, opened 2023-01-20
Conda-installing sbpy causes tests to fail on collection for Python<3.11


astronomerritt:
**UPDATE 2023-03-22:** The issue below occurs _when sbpy is conda-installed_. Nothing to do with pytables. Setting up the conda environment with every package needed, and then pip-installing sbpy, works fine. Leaving this here so nobody misunderstands.

HDF5 output needs the pytables package to work, but in versions of Python<3.11, installing pytables creates a dependency conflict at worst and at best causes pytest to fail to collect every single test with the following numpy warning:

`RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 96 from PyObject`

I have tried this with Python 3.10, 3.9, 3.8 and 3.7. 3.6 works but this causes other issues. 3.11 so far works fine as long as a conda environment is created following the below instructions.

Sorcha itself runs fine despite this issue, possibly because the warnings are triggering in the background.

### To replicate:

Pull the sorcha repo and cd into it.

Create a new conda environment using:

`conda create -n sspp_test -c conda-forge -c mjuric python=3.10 spiceypy=4.0.1 openorb numpy pandas matplotlib spice-utils pip pytest astropy scipy sbpy pytables`

and activate it.

Then run `pip install -e .'.

Then run pytest from inside the repo folder.

### Notes:
- From the traceback, it looks like the problem is with the astropy.modelling functions called by sbpy. astropy.modeling is calling astropy.table, and therein triggers the warning.

astronomerritt:
`/opt/miniconda3/envs/sspp_test_37/lib/python3.7/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1006: in _gcd_import
    ???
<frozen importlib._bootstrap>:983: in _find_and_load
    ???
<frozen importlib._bootstrap>:953: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:219: in _call_with_frames_removed
    ???
<frozen importlib._bootstrap>:1006: in _gcd_import
    ???
<frozen importlib._bootstrap>:983: in _find_and_load
    ???
<frozen importlib._bootstrap>:953: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:219: in _call_with_frames_removed
    ???
<frozen importlib._bootstrap>:1006: in _gcd_import
    ???
[...]

PR #354, opened 2023-01-20
Testing HDF5 output/pytables.

astronomerritt:
I have added pytables to setup.cfg to enable HDF5 output. I expect this will cause github's automated tests to fail due to #353. Currently, all tests run successfully on my machine using Python 3.11.


mschwamb:
@astronomerritt all the tests passed

astronomerritt:
Yep, because the test environments pip install everything - they were failing before because I had the wrong name for pytables in the setup.cfg file. (It's pytables on conda, tables on pip. I hate it when that happens.) Sorry, totally forgot to mention that part in the chaos of trying to fix it.

You can merge this pull request - HDF5 output is working now, and it's being unit tested.

mschwamb:
Is it possible to make these separate pull requests for then other additions?. Because if pytables issue is not resolved in conda then this shouldn't be merged until we know it's not causing issues with install

astronomerritt:
This pull request does not change the installation in any significant way apart from adding pytables to the requirements in setup.cfg, which is needed to make HDF5 output work.

The issue we're having in #353 is with conda installing sbpy, which isn't related to what's included in the repo at all. I've updated the title to reflect that.

(Pytables was relevant at first because adding pytables to my environment is what brought this issue to my attention. However, _pytables isn't the problem_.)

A fresh install works fine on my machine, anyway.

If you still want it separated, close the request and I'll revert on my end - this will however mean that HDF5 output won't work.

mschwamb:
Can a clean install be done with conda and pip installing the package in this pull request (no special pip install of sbpy)? If that is not the case these need to be split.

astronomerritt:
Yes, by following the current instructions in the docs - creating the conda environment, then using pip install -e . to install the SSPP requirements. No special instructions needed.

However, even if I revert this pull request, we will still have the same problem with sbpy. It will always need to be pip installed until we've worked out what's causing the conda installation to generate warnings during tests. This pull request doesn't do anything to alter the problem. It just adds pytables to setup.cfg.

PR #411, opened 2023-07-10
Create generic data readers

jeremykubica:
Create an abstract `ObjectDataReader` class which can be used to load in data (ephemeris or auxiliary data) for objects. This class provides the ability to read files in chunks based on line numbers (for loading the first X rows) or a set of object IDs. This will allow us to load either the ephemeris or auxiliary data in row-based chunks and the other data sets in object ID based chunks.

A `CSVDataReader` is included to read CSV or whitespace separated files. Tests cover both the OIF ephemeris files and the auxiliary files.

This is the first step in the reader refactoring. The readers are not integrated into the actual code yet.

- [X] Does SurveySimPP run successfully on a test set of input files/databases?
- [X] Have you used flake8 on the files you have updated to confirm python programming style guide enforcement? (You can ignore line length warnings: flake8 --ignore=E501,E126,E228,E133,W503)


src/surveySimPP/readers/CSVReader.py
```diff
@@ -0,0 +1,191 @@
...
+        obj_ids (list): A list of object IDs to use.
+
+        Returns:
+        -----------
+        res_df (Pandas dataframe): The dataframe for the ephemerides.
+        """
+        self._build_id_map()
+
+        # Create list of only the matching rows for these object IDs and the header row.
+        row_good = [False] * self.header_row  # pre-header
```
drewoldag, line 173: This is a minor point, feel free to ignore - given the way that the `row_good` list is used in the lambda, could we rename it? Something like the following?

```
skipped_row = [True] * self.header_row  # skip the pre-header
skipped_row.extend([False])  # Keep the the column header
skipped_row.extend(~self.obj_id_table["ObjId"].isin(obj_ids).values)

...

res_df = pd.read_csv( ...
    skiprows=(lambda x: skipped_row[x]),
)
```

drewoldag (approved):
This looks like a great starting point! Thanks for putting this together.

PR #416, opened 2023-07-14
Add HFD5 and OIF Readers

jeremykubica:
This is step three in generalizing the reader code. 
- We add two new reader types `HDF5DataReader` and `OIFDataReader`.  
- It also converts the main code path to use OIFDataReader instead of `PPReadEphemerides` and `PPReadOif`.
- Add the ability to cache the pandas table in-memory for all reader types.
- Refactor the API of the inherited types (by adding internal reader functions for the subclasses) and moving the caching + validation to the baseclass.

There are a few approaches that I would appreciate thoughts on whether there are better approaches:
1) `OIFDataReader` is a reader class that inherits from `ObjectDataReader` but creates an internal object of either `CSVDataReader` and `OIFDataReader`.
2) The logic of `process_and_validate_input_table` is a bit messy with respect to inheritance. Is there a better way to do this?
3) Does the logic for `_read_rows_internal` and `_read_objects_internal` make sense?

- [X] Does SurveySimPP run successfully on a test set of input files/databases?
- [X] Have you used flake8 on the files you have updated to confirm python programming style guide enforcement? (You can ignore line length warnings: flake8 --ignore=E501,E126,E228,E133,W503)


src/surveySimPP/readers/ObjectDataReader.py
```diff
@@ -1,19 +1,42 @@
...
+and returning a pandas data frame.
+
+Each subclass of ObjectDataReader must implement at least the functions
+_read_rows_internal and _read_objects_internal, both of which return a
+pandas data frame. Each data source needs to have a column ObjID that
+identifies the object and can be used for joining and filtering.
+
+Caching is implemented in the base class. This will lazy load the full
+table into memory from the chosen data source, so it should only be
+used with smaller data sets. Both``read_rows`` and ``read_objects``
```
drewoldag, line 11: minor: White space between Both and "``read_rows``.

drewoldag:
Going back and reviewing each of the questions that you've left in the summary for this PR. 

tl;dr; 
It's all good.

The dets;
1. **OIFDataReader is a reader class that inherits from ObjectDataReader but creates an internal object of either CSVDataReader and OIFDataReader.** - I believe that the second internal object is a `HDF5DataReader`. I really like this approach. I'm not a huge fan of requiring the user to provide one of multiple different strings to specify the same thing (i.e. "whitespace" = "comma" = "csv") but that's a minor point.  
2. **The logic of process_and_validate_input_table is a bit messy with respect to inheritance. Is there a better way to do this?** I agree - this is a bit messy. In the case of` `OrbitAuxReader` the data is being processed by 3 classes worth of `_process_and_validate_input_table`, and that makes a lot of sense. However, with the sort of dependency injection going on in `OIFReader.py`, it takes longer to parse through exactly which `_process_and_validate...` methods are being called and when. I think this is just a required complication of dynamically selecting the Reader class, not sure if there's a way around it.   
3. **Does the logic for _read_rows_internal and _read_objects_internal make sense?** - Yes, everything that I've seen here makes sense to me. I think this looks good.

drewoldag (approved):
I left a couple small comments, but overall this looks ok to me. I left a larger commend in the Conversation thread of this PR addressing the questions @jeremykubica left in the PR summary.

src/surveySimPP/readers/CSVReader.py
```diff
@@ -197,6 +161,44 @@ def read_objects(self, obj_ids, **kwargs):
...
+        input_table (Pandas dataframe): Returns the input dataframe modified in-place.
+        """
+        # Perform the parent class's validation (checking object ID column).
+        input_table = super()._process_and_validate_input_table(input_table, **kwargs)
+
+        # Strip out the whitespace from the column names.
+        input_table = input_table.rename(columns=lambda x: x.strip())
+
+        # Check for NaNs or nulls.
+        if "disallow_nan" in kwargs and kwargs["disallow_nan"]:  # pragma: no cover
```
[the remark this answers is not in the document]
jeremykubica: I think the `.get()` is the better approach. I honestly forgot that was an option.

I've changed it in the code.

src/surveySimPP/readers/ObjectDataReader.py
jeremykubica: Fixed.

PR #435, opened 2023-07-19
Final step in converting the data readers

jeremykubica:
Replaces the code path in sorcha.py with the a new `CombinedDataReader` that will do chunked reading of all the input files (ephemeris and auxiliary files). This class provides the ability to read in chunks based on any of the data files, including chunks of the ephemeris file.

Adapts the current test_PPReadAllInput.py to cover the behavior of the new code path.


src/sorcha/readers/CombinedDataReader.py
```diff
@@ -0,0 +1,124 @@
...
+        # No more data to load.
+        if len(obj_ids) == 0:
+            return None
+
+        # Load in the data for this block.
+        if ephem_df is None:
+            ephem_df = self.ephem_reader.read_objects(obj_ids)
+        ephem_ids = set(ephem_df["ObjID"].unique().tolist())
+
+        for reader in self.aux_data_readers:
```
drewoldag, line 109: This for loop isn't actually doing another full read of the first aux_reader, right? I think that it's using the inverse index to read in only the required rows, but I wanted to make sure that my understanding is correct.

src/sorcha/sorcha.py
```diff
@@ -89,11 +94,28 @@ def runLSSTPostProcessing(cmd_args):
...
+        ephem_type = configs["ephemerides_type"]
+        if ephem_type.casefold() != "oif":  # pragma: no cover
+            pplogger.error(f"PPReadAllInput: Unsupported value for ephemerides_type {ephem_type}")
+            sys.exit(f"PPReadAllInput: Unsupported value for ephemerides_type {ephem_type}")
+        reader.add_ephem_reader(OIFDataReader(cmd_args["oifoutput"], configs["eph_format"]))
+
+    reader.add_aux_data_reader(OrbitAuxReader(cmd_args["orbinfile"], configs["aux_format"]))
+    reader.add_aux_data_reader(CSVDataReader(cmd_args["paramsinput"], configs["aux_format"]))
+    if configs["comet_activity"] == "comet":
+        reader.add_aux_data_reader(CSVDataReader(cmd_args["cometinput"], configs["aux_format"]))
```
drewoldag, line 115: This looks so good. One thing that has come up with the light curve variability work is that there's a reasonably good chance that users will create new functions that require parameters beyond what is currently read in. 

This works sets us up so nicely for that.

drewoldag (approved):
This is awesome work. Thanks for putting it together!

src/sorcha/readers/CombinedDataReader.py
jeremykubica: That is correct. Outside the for loop, the program is reading in a set range of rows from `aux_data_readers[0]`. During the first iteration of the for loop, the program will re-open the file and load the rows for the matching object IDs.

That being said, this is still reopening the file and reading it back in. Let me update it to only do that if it has not already loaded the `aux_data_readers[0]` data.

drewoldag (approved):
That change looks good to me.

PR #437, opened 2023-07-19
Implementing difi

astronomerritt:
Most of the code in PPLinkingFilter has now been replaced by a call to difi. Tests have been updated and there is an additional config file variable, SSP_maximum_time, which controls the maximum time between subsequent observations for a tracklet to be formed. Tests have been updated and I've added difi to the pyproject.toml. 

The unit test for PPLinkingFilter is bad, sorry. I am going to replace it at the same time I make the demo notebook.


pyproject.toml
```diff
@@ -28,6 +28,7 @@ dependencies = [
     "matplotlib",
     "sbpy",
     "tables",
+    "difi"
```
moeyensj, line 31: I released a pre-release which apparently by default doesn't get installed with `pip install difi` so I think pinning the version here will fix the import issues. 
```suggestion
    "difi == 1.2rc1"
```

PR #443, opened 2023-07-19
Adding configuration files to simulate full survey runs

### LINKED ISSUE #389: PPConfig example file for Rubin
It would be very helpful to have a PPConfig.ini file, ideally the demo file, that is as Rubin-like as possible. The current file has some major differences from Rubin (observing_filters doesn't include u or y), and also includes enough settings that telescope non-experts (like me) might find it difficult to determine what is already set to Rubin baseline and what is not. 

Personally, I just did a bunch of work with a version of the PPConfig file I'd modified from the demo, and only realized late in the process that I needed to change the observing_filters setting.

fedorets:
Fixes #389 

Created survey_setups folder to store configuration files simulating the surveys.

Two configuration files (Rubin_full_footprint.ini and Rubin_circular_approximation.ini) represent the two different use cases (respectively, the full footprint and circular approximation).

### DIFF
--- a/survey_setups/README
+++ b/survey_setups/README
@@ -1,0 +1,17 @@
+README file for survey_setups folder
+
+This directory contains configuration files that simulate Solar System surveys.
+
+The user are encouraged to familiarise themselves with the documentation for 
+a deeper understanding of different parameters. 
+
+Currently, the directory contains the following files:
+
+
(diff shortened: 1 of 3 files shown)


mschwamb (changes requested):
Please add a read.me to explain what these two files are and people can add to them

mschwamb (approved):

PR #446, opened 2023-07-19
Implementing difi

astronomerritt:
Most of the code in PPLinkingFilter has now been replaced by a call to difi. Tests have been updated and there is an additional config file variable, SSP_maximum_time, which controls the maximum time between subsequent observations for a tracklet to be formed. Tests have been updated and I've added difi to the pyproject.toml - this time specifying the pre-release version as suggested by @moeyensj, so hopefully it builds with the correct version and the checks don't fail...


src/sorcha/modules/PPLinkingFilter.py
```diff
@@ -1,9 +1,6 @@
 import pandas as pd
 import numpy as np
-from astropy import units as u
-from astropy.coordinates import SkyCoord
-
-from .PPDetectionEfficiency import PPDetectionEfficiency
+from difi.metrics import NightlyLinkagesMetric
 
 
 def PPLinkingFilter(
```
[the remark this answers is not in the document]
astronomerritt: It's run serially on chunks of multiple objects. There's no multiprocessing currently in Sorcha at all - I don't know if there's any plans to change this.

src/sorcha/modules/PPLinkingFilter.py
```diff
@@ -49,105 +48,50 @@ def PPLinkingFilter(
...
 
-    # flatten the final index
-    flat_idx = [x for xs in final_idx for x in xs]
+    findable, window_summary = metric.run(
+        difi_dataframe,
+        detection_window=tracklet_interval,  # number of nights in a rolling detection window
+        min_window_nights=min_tracklets,  # minimum size of a window as the rolling window is moved
+        by_object=True,  # split observations by object instead of by window [set to True if you want ignore_after_discovery=True]
+        discovery_probability=detection_efficiency,  # probability of discovery for any discovery chance
+        num_jobs=30,  # number of parallel jobs
```
moeyensj, line 91: If num_jobs is set to None, it will automatically use all cores available on the system. But depending on how this is used in the simulator this might also need to be 1.

src/sorcha/modules/PPLinkingFilter.py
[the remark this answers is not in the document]
moeyensj: I'll file an issue against difi to address the multiprocessing concern.

src/sorcha/modules/PPLinkingFilter.py
```diff
@@ -49,105 +48,50 @@ def PPLinkingFilter(
...
+    metric = NightlyLinkagesMetric(
+        linkage_min_obs=min_observations,  # min observations per tracklet
+        max_obs_separation=maximum_time,  # maximum time separation between subsequent observations in a tracklet [days]
+        min_linkage_nights=min_tracklets,  # minimum number of nights on which a tracklet must be detected
+        min_obs_angular_separation=minimum_separation,  # minimum angular separation between observations in a tracklet [arcseconds]
+    )
 
-    # flatten the final index
-    flat_idx = [x for xs in final_idx for x in xs]
+    findable, window_summary = metric.run(
```
drewoldag, line 85: Minor point: I don't _think_ that we're using `window_summary` after this. If that's true then we can replace it with `_`, i.e. `findable, _ = metric.run(...)`, to avoid creating the new variable. But if `window_summary` is, or will be, used, then please ignore this comment. :)

src/sorcha/modules/PPLinkingFilter.py
drewoldag: @moeyensj if `num_jobs` is set to a value larger than the number of cores available, will `difi` just use the core that are available? Should this be made into a config parameter?

drewoldag (approved):
Overall this looks great. I left a couple of small comments, but it would be fine to convert them into new issues and merge this branch.

PR #448, opened 2023-07-20
Lightcurve

### LINKED ISSUE #143: adding asteroid physical parameters (shapes, rotation)
Hi, I'm currently at the LSSC workshop at the CCA and have been speaking with Mario about adding some lightcurve data (shape, rotation, etc) to this post-processing pipeline. It would be great to talk about how I can get involved with this project. Thanks!

bernardinelli:
Implements the structure for light curve variability for all objects within the simulation. Currently, only a sinusoidal light curve is implemented - more to come soon!

fixes #143 
 

demo/OIFconfig_benchmark.ini
```diff
@@ -150,4 +150,8 @@ magnitude_decimals = 3
 # They may have unexpected results or break the code entirely.
 
 # SQL query for extracting data from the pointing database.
-pointing_sql_query = SELECT observationId, observationStartMJD, filter, seeingFwhmGeom, seeingFwhmEff, fiveSigmaDepth, fieldRA, fieldDec, rotSkyPos FROM observations order by observationId
\ No newline at end of file
+pointing_sql_query = SELECT observationId, observationStartMJD, filter, seeingFwhmGeom, seeingFwhmEff, fiveSigmaDepth, fieldRA, fieldDec, rotSkyPos FROM observations order by observationId
+
+
+lightcurve = False
+lc_model = Sinusoidal
```
drewoldag, line 157: minor point - I wouldn't be shy about using a longer config name here. i.e. `lightcurve_model` would be fine in my book :)

demo/OIFconfig_lca.ini
```diff
@@ -0,0 +1,157 @@
...
+
+# WARNING: DO NOT CHANGE THESE PARAMETERS UNLESS YOU ARE VERY SURE OF WHAT YOU ARE DOING!
+# They may have unexpected results or break the code entirely.
+
+# SQL query for extracting data from the pointing database.
+pointing_sql_query = SELECT observationId, observationStartMJD, filter, seeingFwhmGeom, seeingFwhmEff, fiveSigmaDepth, fieldRA, fieldDec, rotSkyPos FROM observations order by observationId
+
+
+lightcurve = True
+lc_model = Sinusoidal
```
drewoldag, line 157: Same comment as in OIFconfig_bencmark.ini

demo/bench_LCA.py
```diff
@@ -0,0 +1,35 @@
+# A basic script Jeremy hacked together to run cProfile on a subset of
+# 1000 TNOs.
```
drewoldag, line 2: Please update the comment here :)

demo/bench_LCA.py
```diff
@@ -0,0 +1,35 @@
...
+        'makeTemporaryEphemerisDatabase': False,
+        'readTemporaryEphemerisDatabase': False,
+        'deleteTemporaryEphemerisDatabase': False,
+        'surveyname': 'LSST',
+        'outfilestem': f"out_{args.object_type}",
+        'verbose': False,
+    }
+
+    cProfile.run("runLSSTPostProcessing(cmd_args_dict)", "../data/out/restats")
+    # runLSSTPostProcessing(cmd_args_dict)
```
drewoldag, line 32: I think we can remove line 32, yeah?

src/sorcha/modules/PPConfigParser.py
```diff
@@ -639,6 +639,27 @@ def PPConfigFileParser(configfile, survey_name):
...
+    except ValueError:
+        pplogger.error(
+            "ERROR: could not parse value for lightcurve as a boolean. Check formatting and try again."
+        )
+        sys.exit(
+            "ERROR: could not parse value for lightcurve as a boolean. Check formatting and try again."
+        )
+
+    config_dict["lc_model"] = config.get("EXPERT", "lc_model", fallback=None)
+    config_dict["lc_model"] = None if config_dict["lc_model"] == "None" else config_dict["lc_model"]
```
drewoldag, line 653: I don't think I understand what this line is doing. Is the logic here necessary only because `config_dict["lc_model"]` can have the string value "None" instead of the python `None`?

src/sorcha/modules/PPConfigParser.py
```diff
@@ -639,6 +639,27 @@ def PPConfigFileParser(configfile, survey_name):
...
+
+    if config_dict["lightcurve"] and not config_dict["lc_model"]:
```
drewoldag, line 655: Would it make sense to simplify this logic? i.e. `config_dict['lc_model'] == None` implies that `config_dict['lightcurve'] == False`, otherwise True?

src/sorcha/utilities/baseLightCurve.py
drewoldag, line 1: I kind of get the feeling that this could blossom into a much larger set of files potentially. For instance, if the _compute expands to make calls to multiple other functions modeling the variety of different things that can affect the light curve, then those functions might be kept in distinct modules next to this one. 

I would advocate for moving this into a new folder, perhaps `.../src/lightcurves/` (name TBD). Additionally, I would advocate for moving the `SinusoidalLightCurve` class into its own python module, and letting `BaseLightCurve` be the only class in this module.

src/sorcha/utilities/baseLightCurve.py
```diff
@@ -0,0 +1,62 @@
+from abc import ABC, abstractmethod
+import numpy as np 
+
+timeColumn = 'FieldMJD'
+
+class baseLightCurve(ABC):
```
drewoldag, line 6: Do we expect anyone to ever actually use the `BaseLightCurve` class for doing work? If not, perhaps it would make sense to rename this to something like `AbstractLightCurve`?

If that makes sense, you could also replace lines 29-31 with lines 56-61.

drewoldag (approved):
This looks like a great starting point to me. I've left a few questions, and I'm happy to chat with you in person about them. I get the feeling that some of them are probably just misunderstanding on my part :)

PR #456, opened 2023-07-20
Making changes to config variables.

### LINKED ISSUE #439: Don't have multiple options in the config file that do the same thing
eph_format can accept "comma" and "csv" as options, but both of these do the same thing. We should remove "comma".

### LINKED ISSUE #442: Change output format from "default" to "basic"
The options for output format are currently "default" and "all". "Basic" would be a better label.

astronomerritt:
Fixes #439.
Fixes #442.

I have removed the "comma" option from the aux_format as it is completely superfluous and people should know what a csv is. This change has been carried right through to /src/sorcha/readers/CSVReader.py and all tests updated.

I have also renamed the "default" output option to the much clearer "basic" and again updated tests. Also made a small change so that the code explicitly checks for "all" before writing out all columns. This isn't really necessary (as the variable can only be "basic" or "all", checked at config parsing) but will be clearer if we add more options in the future.


maxwest-uw (approved):
no conflicts with my changes from what I can see

PR #465, opened 2023-07-25
Refactored `Footprint` object creation

### LINKED ISSUE #464: Refactor detector `Footprint` instantiation
Currently the detector footprint is created each time `PPApplyFOVFilter` is called - twice per chunk. The creation of the `Footprint` object is a rather inefficient list comprehension over the array of (x,y) pairs (roughly speaking, it's a nested for-loop) that are read in from a CSV that does not change while Sorcha is running. 

So in this case, we can optimize the generation of the `Footprint` object by creating a temporary dictionary of `detector_id : indicies` after reading in the CSV file, and use that temporary dictionary to optimize the way the list of `Detector` objects is created. i.e. instead of looping over the entire pandas dataframe looking for matching values, we just extract the entries we need with the indicies from the temp. dictionary. 

Additionally, since the detector (x,y) pairs does not change during the Sorcha, run, we can create the `Footprint` object once, before the first chunk is read in, and then reuse that for the duration of the execution.

drewoldag:
Moved `Footprint` instantiation out of `PPApplyFOVFilter`. Set difi `num_jobs=1`.

Fixes #464  .

Please see the ticket for a description of the what the code was originally organized. This PR represents the work to move the creation of the `Footprint` object (that describes the detector geometry) out of `PPApplyFOVFilter` (which is called twice per chunk) into the main sorcha function `runLSSTPostProcessing` such that it only needs to be calculated once. 

Additionally, I modified the list comprehension step. When a `Footprint` object is being instantiated it reads in a CSV file containing a list of (detector id, x, y) tuples. A nested loop was being used to generate a list of `Detector` objects. I removed the nested loop in favor of direct indexing. 

I used the `.../demo/bench_cProfile.py` script to benchmark the changes.  I've summarized the changes below: 

| Change | Calls | PPFootprintFilter time | Benchmark run time |
| ---- | ---- | ---- | ---- | 
| Baseline | 20 | 2.464s | 11.878s |
| Direct indexing during instantiation | 20 | 2.184s | 11.499s | 
| Instantiate only 1 time | 1 | 0.111s | 9.438s |

### DIFF
--- a/src/sorcha/modules/PPLinkingFilter.py
+++ b/src/sorcha/modules/PPLinkingFilter.py
@@ -88,7 +88,7 @@ def PPLinkingFilter(
         min_window_nights=min_tracklets,  # minimum size of a window as the rolling window is moved
         by_object=True,  # split observations by object instead of by window [set to True if you want ignore_after_discovery=True]
         discovery_probability=detection_efficiency,  # probability of discovery for any discovery chance
-        num_jobs=30,  # number of parallel jobs
+        num_jobs=1,  # number of parallel jobs
         ignore_after_discovery=True,  # ignore observations of an object after it has been discovered
     )
 
--- a/src/sorcha/modules/PPApplyFOVFilter.py
+++ b/src/sorcha/modules/PPApplyFOVFilter.py
@@ -1,10 +1,9 @@
-from . import PPFootprintFilter
 import logging
 import numpy as np
 from astropy.coordinates import SkyCoord
 
 
-def PPApplyFOVFilter(observations, configs, rng, verbose=False):
+def PPApplyFOVFilter(observations, configs, rng, footprint=None, verbose=False):
     """
     Wrapper function for PPFootprintFilter and PPFilterDetectionEfficiency. Checks to see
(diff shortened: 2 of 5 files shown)


src/sorcha/modules/PPLinkingFilter.py
```diff
@@ -88,7 +88,7 @@ def PPLinkingFilter(
         min_window_nights=min_tracklets,  # minimum size of a window as the rolling window is moved
         by_object=True,  # split observations by object instead of by window [set to True if you want ignore_after_discovery=True]
         discovery_probability=detection_efficiency,  # probability of discovery for any discovery chance
-        num_jobs=30,  # number of parallel jobs
+        num_jobs=1,  # number of parallel jobs
```
drewoldag, line 91: Note that I set this to 1. @maxwest-uw noted that there seemed to be quite a lot of overhead associated with values other than 1, so to avoid that, and until we can characterize that penalty more clearly, I suggest that we keep this as 1 for now.

astronomerritt (approved):
Looks good to me!

jeremykubica (approved):

PR #475, opened 2023-07-28
Updating difi version in pyproject.toml.

### LINKED ISSUE #476: Difi triggering FileExistsError: [Errno 17] File exists: '/DIFI_ARRAY' error
Today I found quite suddenly that any calls to difi resulted in a `FileExistsError: [Errno 17] File exists: '/DIFI_ARRAY' `error which persisted even in a fresh environment. (I'll post the full traceback below.)

Installing v.2rc3 stopped this as now shared memory arrays have an instance-specific name including the PID. So I think it's a good idea to update the requirement for difi to this version in the pyproject.toml file.

I'm not sure why this happened (looks like a file isn't being cleared from memory when it should be?) but I assume that if it happened to me it could happen to someone else.

astronomerritt:
Fixes #476.

I've updated the difi requirement in pyproject.toml to the latest version, v1.2rc3. 

This is specifically because I found quite suddenly that any calls to difi resulted in a `FileExistsError: [Errno 17] File exists: '/DIFI_ARRAY'` error which persisted even in a fresh environment. Installing v.2rc3 stopped this as now shared memory arrays have an instance-specific name including the PID.

I'm not sure why this suddenly started happening on my machine (explanations and speculations welcome) but I figured to be safe it's best to update the version requirement.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ dependencies = [
     "matplotlib",
     "sbpy",
     "tables",
-    "difi == 1.2rc2",
+    "difi == 1.2rc3",
 ]
 
 [project.scripts]


drewoldag (approved):
Approved!

maxwest-uw (approved):

PR #478, opened 2023-07-28
Fixing error with cometary case

### LINKED ISSUE #431: Cometary case fails with correctly formatted data
When running a test case with cometary parameters involved, although all input is correct, the programme fails with the 'ERROR: no comet activity specified.' error.

The issue has to do with the log parsing as the used parameter in the config parser  is mismatched. The error should be solved by changing line 684 in the modules/PPConfigParser.py from 

        pplogger.info('Cometary activity set to: ' + str(configs['cometary activity']))
 
       to
 
        pplogger.info('Cometary activity set to: ' + str(configs['comet_activity']))

### LINKED ISSUE #477: Cometary case does not calculate distance and phase angle offset
When calculating magnitudes with the flag comet_activity = comet, the calculated magnitude is H + colour + phase angle, with the distance and phase angle effects omitted.

fedorets:
Fixes #477 (and also the recently closed #431 , which was not properly updated last time)

- Added the distance offset and phase angle calculation when the cometary case is switched on.
- Modified the relevant test accordingly.
- Fixed one instance in the config parser which was still wrong

### DIFF
--- a/src/sorcha/modules/PPCalculateApparentMagnitude.py
+++ b/src/sorcha/modules/PPCalculateApparentMagnitude.py
@@ -50,6 +50,7 @@ def PPCalculateApparentMagnitude(
     if object_type == "comet":
         verboselog("Calculating cometary magnitude using a simple model and applying colour offset...")
         observations = PPCalculateSimpleCometaryMagnitude(observations, mainfilter, othercolours)
+        observations.rename(columns={"H_" + mainfilter: "H_filter"}, inplace=True)
     else:
         # if user is only interested in one filter, we have no colour offsets to apply: assume H is in that filter
         if len(observing_filters) > 1:
@@ -60,9 +61,9 @@ def PPCalculateApparentMagnitude(
         else:
             observations.rename(columns={"H_" + mainfilter: "H_filter"}, inplace=True)
 
-        verboselog("Calculating apparent magnitude in filter...")
-        observations = PPCalculateApparentMagnitudeInFilter(
-            observations, phasefunction, lightcurve=lightcurve, lightcurve_choice=lightcurve_choice
-        )
+    verboselog("Calculating apparent magnitude in filter...")
+    observations = PPCalculateApparentMagnitudeInFilter(
+        observations, phasefunction, lightcurve=lightcurve, lightcurve_choice=lightcurve_choice
+    )
 
     return observations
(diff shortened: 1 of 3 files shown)


src/sorcha/modules/PPCalculateApparentMagnitude.py
```diff
@@ -60,9 +61,9 @@ def PPCalculateApparentMagnitude(
         else:
             observations.rename(columns={"H_" + mainfilter: "H_filter"}, inplace=True)
 
-        verboselog("Calculating apparent magnitude in filter...")
-        observations = PPCalculateApparentMagnitudeInFilter(
-            observations, phasefunction, lightcurve=lightcurve, lightcurve_choice=lightcurve_choice
-        )
+    verboselog("Calculating apparent magnitude in filter...")
+    observations = PPCalculateApparentMagnitudeInFilter(
+        observations, phasefunction, lightcurve=lightcurve, lightcurve_choice=lightcurve_choice
```
mschwamb, line 66: I am not sure if we should be applying a cometary phase curve or if this is already in the comet class. Let's get some configuration and a calculation of a comet mag at different distances from Colin before this pull request is evaluated

mschwamb: @astronomerritt can follow-up on this

astronomerritt: I've emailed Colin.

astronomerritt: However: @fedorets I see what's been done here, but this wouldn't fix the issue by itself even if this is the way we want to do it. Note that PPCalculateSimpleCometaryMagnitude creates the TrailedSourceMag column by default. PPCalculateApparentMagnitudeInFilter operates on H_filter and also creates TrailedSourceMag, overwriting what was done by PPCalculateSimpleCometaryMagnitude. 

This could be easily fixed by changing the default colname in PPCalculateSimpleCometaryMagnitude.py to "H_filter" (line 8). Then you could remove the line 53 you've added here in PPCalculateCometaryMagnitude.

astronomerritt:
I'm closing this pull request and recreating it - Grigori is on vacation and while this is very nearly right (thanks Grigori!), there's some further alterations to be made for it to work properly. Simpler for me to implement these changes myself in my upcoming pull request. Also avoids the merge conflict :)

Issue #486, opened 2023-07-28
Warn the user if the requested lightcurve model isn't found in `LC_METHODS`


drewoldag:
If the user requests a lightcurve model that is not present in the LC_METHODS dictionary, we'll fail gracefully and continue on in the code. 

We should change that in one of two ways. Either expand the conditional in src/sorcha/modules/PPCalculateApparentMagnitudeInFilter.py so that there is some logging that is presented to the user. Or expand the logic in config parser script such that it will check for the existence of the lightcurve method before any calculations are made.

mschwamb:
I think this check should happen in the config parser. Especially if the simulation is asking sorcha to run the ephemeris generation it could take a bit to get to the apparent magnitude function, and it would be better to fail and return an error early for issues with settings/configurations if possible

mschwamb:
Also ditto for cometary activity model

Issue #482, opened 2023-07-28
Remove `lightcurve` boolean configuration parameter


drewoldag:
There isn't a need for both a `lightcurve:bool` and a `lc_model:str` parameter in the configuration file. 

We should remove the boolean and assume that if the user passes `"none"`, `""`, or and unknown light curve model name, that we will not execute lightcurve model calculations.

mschwamb:
Good catch. That was the original intention, but I might have missed in the approval that this wasn't the case

PR #487, opened 2023-07-28
Prefactor of lightcurve code

drewoldag:
Initial commit to refactor lightcurve code in prep for separating it out to `sorcha_community_util` repo.

Fixes # 484.

This PR represents the work to:

1. Separate the abstract base class for the lightcurve models from any implementations
2. Introduces a test LC model "identity"
3. Updates docstrings for the base and identity classes
4. Removes the use of the `lightcurve` boolean config value (but does not remove it from the config machinery yet (there's a separate issue for that #482 )
5. Adds tests for the lightcurve registration mechanism
6. Expands the `.../lightcurves/init.py` file so that the classes are registered automatically.
7. Update `src/sorcha/modules/PPCalculateApparentMagnitudeInFilter.py` where we make use of LC_METHODS so that if the requested light curve model isn't in the dictionary, the program won't crash (extra work needed here to report something reasonable to the user when this happens #486)

Bonus: 
Add the extra config parameter needed in the `demo/OIFconfig_*` files so that `bench_cProfile` runs correctly. 

### DIFF
--- a/src/sorcha/lightcurves/base_lightcurve.py
+++ b/src/sorcha/lightcurves/base_lightcurve.py
@@ -1,4 +1,5 @@
 from abc import ABC, abstractmethod
+from typing import List
 import numpy as np
 import pandas as pd
 
@@ -6,7 +7,7 @@ import pandas as pd
 class AbstractLightCurve(ABC):
     """Abstract base class for lightcurve models"""
 
-    def __init__(self, required_column_names: list[str] = []) -> None:
+    def __init__(self, required_column_names: List[str] = []) -> None:
         """Abstract base class accepts a list of column names that are required
         to be present in the Pandas dataframe passed to the ``compute`` method.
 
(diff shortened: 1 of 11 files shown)


bernardinelli (approved):
This looks good to me! (I also got lost in the config variables to make them one instead of two...)

src/sorcha/lightcurves/base_lightcurve.py
```diff
@@ -1,55 +1,78 @@
...
     @abstractmethod
-    def name_id():
+    def name_id() -> str:
         """This method will return the unique name of the LightCurve Model"""
-        raise (NotImplementedError, "Must be implemented as a static method by the child class")
+        raise (NotImplementedError, "Must be implemented as a static method by the subclass")
+
+
+# ! The code below will be moved to the `sorcha_community_utils` repository soon
+TIME_COLUMN = "FieldMJD"
```
maxwest-uw, line 55: this doesn't need to be changed this pr because it's going to be moved, but the time column shouldn't be a hard coded constant if we want to have support for other surveys or for other object generation code (i.e., if we decide to get rid of OIF eventually)

maxwest-uw (approved):
I'm approving, with the additional request that we make the `TIME_COLUMN` into not a constant in a future PR/when we move it the `sorcha_community_utils` module :)

src/sorcha/lightcurves/base_lightcurve.py
drewoldag: Agreed. I'll go ahead and remove it here just so we don't forget about it. I'll make the same change in the `IdentityLightcurve` class.

drewoldag:
@maxwest-uw I went ahead and made the update that you suggested so that it isn't forgotten. 😃

PR #493, opened 2023-07-31
Remove unused `lightcurve` config

### LINKED ISSUE #482: Remove `lightcurve` boolean configuration parameter

drewoldag:
Fixes #482 .

This PR removes the unused `lightcurve` config parameter and the various references to it throughout the code and tests. 

### DIFF
--- a/src/sorcha/modules/PPCalculateApparentMagnitudeInFilter.py
+++ b/src/sorcha/modules/PPCalculateApparentMagnitudeInFilter.py
@@ -8,7 +8,7 @@ from sorcha.lightcurves.lightcurve_registration import LC_METHODS
 
 
 def PPCalculateApparentMagnitudeInFilter(
-    padain, function, colname="TrailedSourceMag", lightcurve_choice="None"
+    padain, function, colname="TrailedSourceMag", lightcurve_choice=None
 ):
     """
     This task calculates the apparent brightness of an object at a given pointing
@@ -33,9 +33,7 @@ def PPCalculateApparentMagnitudeInFilter(
 
     colname (string): column name in which to store calculated magnitude.
 
-    lightcurve (boolean): whether lightcurves are applied or not
-
-    lightcurve_choice (string): choice of lightcurve model
+    lightcurve_choice (string): choice of lightcurve model. Default None
 
 
     Returns:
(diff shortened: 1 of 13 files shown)


src/sorcha/modules/PPCalculateApparentMagnitudeInFilter.py
```diff
@@ -8,7 +8,7 @@
 
 
 def PPCalculateApparentMagnitudeInFilter(
-    padain, function, colname="TrailedSourceMag", lightcurve_choice="None"
+    padain, function, colname="TrailedSourceMag", lightcurve_choice=None
```
drewoldag, line 11: The default value seemed to bounce around a bit between where the config value was defined in the .ini file and where it finally is used. 

In PPConfigParser, if the .ini file defines this parameter as the string `"none"` it will be cast to the Python `None`. But then here we were defining the default value to be the string `"none"` again. So I've updated the default value to be `None` to be in alignment with what `PPConfigParser` will output.

src/sorcha/modules/PPCalculateApparentMagnitudeInFilter.py
```diff
@@ -66,7 +64,7 @@ def PPCalculateApparentMagnitudeInFilter(
     alpha = padain["Sun-Ast-Obs(deg)"].values
     H = padain[H_col].values
 
-    if LC_METHODS.get(lightcurve_choice, False):
+    if lightcurve_choice and LC_METHODS.get(lightcurve_choice, False):
```
drewoldag, line 67: Extra security logic here. Since it would be possible for someone to define a subclass of `AbstractLightcurve` with a `name_id` method that returns `None`, and since `None` is a valid key in a python dictionary, we wanted to avoid any potential confusion. 

So if the default value of `lightcurve_choice` is passed in, (i.e. `None`) we will not execute this code - even if someone registers a lightcurve class with the name id `None`.

astronomerritt (approved):
This looks good to me -- all the necessary files have been changed appropriately.

jeremykubica (approved):

PR #502, opened 2023-08-02
Comet corrections, #485, #491

### LINKED ISSUE #485: use the astropy units and quantities for the value for au in km in PPCalculateApparentMagnitudeinFilter
Looking at  PPCalculateApparentMagnitudeinFilter, I see there's a conversion from km to au and it's using a long number to convert from km to au. Reading this one has to check what value this is. Astropy has a units and quantities and it seems better to use this rather than have random divide by 1.495978707e8 in the code 

One example:

https://github.com/dirac-institute/sorcha/blob/8145f00930bfc409e53e7d61a37e72ee2e7db65b/src/sorcha/modules/PPCalculateApparentMagnitudeInFilter.py#L52C1-L52C54

There are two other lines in that function that also divide by 1.495978707e8  

e.g 

>>> import astropy 
>>> from astropy import units as u
>>> x=1.0*u.au
>>> x.to(u.km)
<Quantity 1.49597871e+08 km>
>>> x.to(u.km).value
149597870.7

### LINKED ISSUE #491: Update calculation of cometary contribution to apparent mag to match light curve implementation
Coma contribution to the apparent magnitude should be calculated within PPCalculateApparentMagnitudeInFilter, similarly to how the light curve contribution is included there - potentially these should both be added on at the end. 

This will ease the adjustment where coma contribution is rolled into the new light curve functionality.

astronomerritt:
Fixes #491.
Fixes #485.

- PPCalculateSimpleCometaryMagnitude is now called within PPCalculateApparentMagnitudeInFilter, at the end.
- It now correctly combines the coma apparent magnitude with the "nucleus" apparent magnitude, rather than the absolute magnitude.
- The aperture used to calculate the coma magnitude is now the seeing (seeingFwhmEff).
- H_mainfilter is now preserved in the dataframe instead of overwritten in PPApplyColourOffsets.
- I renamed rho and delta in PPCalculateApparentMagnitudeInFilter to be consistent with usual usage.
- Unit tests updated. test_PPCalculateSimpleCometaryMagnitude now checks _two_ distances for a known comet (67P).
- Demo notebooks updated.
- PPCalculateApparentMagnitudeInFilter now uses astropy units to perform conversion from km to au, instead of hardcoded numbers.

### DIFF
--- a/src/sorcha/modules/PPCalculateApparentMagnitudeInFilter.py
+++ b/src/sorcha/modules/PPCalculateApparentMagnitudeInFilter.py
@@ -55,18 +55,22 @@ def PPCalculateApparentMagnitudeInFilter(
 
     # first, get H, rho, delta and alpha as ndarrays
     # delta, rho and alpha are converted to au from kilometres
-    delta = padain["AstRange(km)"].values / 1.495978707e8
+    delta = (padain["AstRange(km)"].values * u.km).to(u.au).value
 
     try:
-        rho = padain["Ast-Sun(km)"] / 1.495978707e8
+        rho = (padain["Ast-Sun(km)"].values * u.km).to(u.au).value
     except KeyError:
         rho = (
-            np.sqrt(
-                padain["Ast-Sun(J2000x)(km)"].values ** 2
-                + padain["Ast-Sun(J2000y)(km)"].values ** 2
-                + padain["Ast-Sun(J2000z)(km)"].values ** 2
+            (
+                np.sqrt(
+                    padain["Ast-Sun(J2000x)(km)"].values ** 2
+                    + padain["Ast-Sun(J2000y)(km)"].values ** 2
+                    + padain["Ast-Sun(J2000z)(km)"].values ** 2
+                )
+                * u.km
             )
-            / 1.495978707e8
+            .to(u.au)
+            .value
         )
 
     alpha = padain["Sun-Ast-Obs(deg)"].values
--- a/src/sorcha/modules/PPCalculateApparentMagnitudeInFilter.py
+++ b/src/sorcha/modules/PPCalculateApparentMagnitudeInFilter.py
@@ -5,10 +5,17 @@ from sbpy.photometry import HG, HG1G2, HG12_Pen16, LinearPhaseFunc
 import logging
 
 from sorcha.lightcurves.lightcurve_registration import LC_METHODS
+from .PPCalculateSimpleCometaryMagnitude import PPCalculateSimpleCometaryMagnitude
 
 
 def PPCalculateApparentMagnitudeInFilter(
-    padain, function, colname="TrailedSourceMag", lightcurve_choice=None
+    padain,
+    function,
+    mainfilter,
+    observing_filters,
+    colname="TrailedSourceMag",
+    lightcurve_choice=None,
+    object_type="None",
 ):
     """
     This task calculates the apparent brightness of an object at a given pointing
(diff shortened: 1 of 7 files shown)


astronomerritt:
Build is failing because I forgot to edit one of the demo notebooks, fixing now.

src/sorcha/modules/PPCalculateApparentMagnitudeInFilter.py
```diff
@@ -46,14 +53,14 @@ def PPCalculateApparentMagnitudeInFilter(
 
     H_col = "H_filter"
 
-    # first, get H, r, delta and alpha as ndarrays
-    # r, delta and alpha are converted to au from kilometres
-    r = padain["AstRange(km)"].values / 1.495978707e8
+    # first, get H, rho, delta and alpha as ndarrays
+    # delta, rho and alpha are converted to au from kilometres
+    delta = padain["AstRange(km)"].values / 1.495978707e8
```
mschwamb, line 58: Since you're modifying this anyway, can we get rid of these hardcoded numbers?

src/sorcha/modules/PPCalculateApparentMagnitudeInFilter.py
```diff
@@ -46,14 +53,14 @@ def PPCalculateApparentMagnitudeInFilter(
...
 
     try:
-        delta = padain["Ast-Sun(km)"] / 1.495978707e8
+        rho = padain["Ast-Sun(km)"] / 1.495978707e8
     except KeyError:
-        delta = (
+        rho = (
             np.sqrt(
                 padain["Ast-Sun(J2000x)(km)"].values ** 2
                 + padain["Ast-Sun(J2000y)(km)"].values ** 2
```
mschwamb, line 66: Same thing. Can we get rid of the hardcoded numbers to handle #485

mschwamb (changes requested):

src/sorcha/modules/PPCalculateApparentMagnitudeInFilter.py
astronomerritt: Absolutely. I was going to do it in a separate PR, but I can do it now no problem.

mschwamb: Might as well do it in one go.

astronomerritt: Done.

src/sorcha/modules/PPCalculateApparentMagnitudeInFilter.py
astronomerritt: Done.

mschwamb:
Looks fine. There isn't a #502 ticket. Not sure if that's a pull request reference

src/sorcha/modules/PPCalculateApparentMagnitudeInFilter.py
```diff
@@ -46,27 +53,31 @@ def PPCalculateApparentMagnitudeInFilter(
...
-            / 1.495978707e8
+            .to(u.au)
+            .value
         )
 
     alpha = padain["Sun-Ast-Obs(deg)"].values
     H = padain[H_col].values
 
     # calculating light curve offset
-    if lightcurve_choice and LC_METHODS.get(lightcurve_choice, False):
```
drewoldag, line 69: @astronomerritt was removing this part of the conditional an intentional change? I had just recently added it, but I'm curious if I missed an edge case when checking the code.

mschwamb: I don't think it was intentional. I missed the lightcurve_choice was taken out. Steph was moving stuff around with light curves that I told them to put back and let Pedro update. So I think it happened there.

astronomerritt: Oh goodness. I think I may have messed up slightly when merging your changes into my branch. My apologies, and thank you for catching it! Please do feel free to change it back, or I can do it tomorrow.

mschwamb: @drewoldag can you add it back in?

drewoldag: Yes, of course. Not a problem at all :) When there are lots of moving pieces, it can be hard to keep everything organized.

Issue #457, opened 2023-07-20
Refactoring lsstcomet into abstractlightcurve-ish class


mschwamb:
Refactoring lsstcomet into abstractlightcurve-ish class

mschwamb:
I've read through the lsstcomet class and In know the bits that are relevant we're using which is basically calculating the coma magnitude. Happy to hop on a call with you @drewoldag to pair code this.

drewoldag:
A related issue is here: #501

PR #504, opened 2023-08-02
Initial scaffolding commit for plugable cometary activity models.

drewoldag:
Foundational work to address #457  .

This PR introduces nearly identical scaffolding to support user provided, pluggable, cometary activity models. Here we have an abstract base class as well as a "identity" class for testing that returns the input dataframe unchanged. Additionally, the registration utility has been implemented as have tests for that. 

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -105,7 +105,6 @@ write_to = "src/sorcha/_version.py"
 [tool.pytest.ini_options]
 testpaths = [
     "tests",
-    "src/sorcha",
 ]
 
 [tool.black]
--- a/src/sorcha/activity/base_activity.py
+++ b/src/sorcha/activity/base_activity.py
@@ -1,23 +1,16 @@
 from abc import ABC, abstractmethod
-from typing import List
-import numpy as np
 import pandas as pd
+import numpy as np
 
 
-class AbstractLightCurve(ABC):
-    """Abstract base class for lightcurve models"""
+class AbstractCometaryActivity(ABC):
+    """Abstract base class for cometary activity models"""
 
-    def __init__(self, required_column_names: List[str] = []) -> None:
+    def __init__(self) -> None:
         """Abstract base class accepts a list of column names that are required
         to be present in the Pandas dataframe passed to the ``compute`` method.
-
-        Parameters
-        ----------
-        required_column_names : list, optional
-            The list of columns in ``observations`` dataframe that are required
-            for the model calculation, by default []
         """
-        self.required_column_names = required_column_names
+        self.required_column_names = []
 
     @abstractmethod
     def compute(self, df: pd.DataFrame) -> np.array:
--- a/src/sorcha/activity/identity_activity.py
+++ b/src/sorcha/activity/identity_activity.py
@@ -1,0 +1,55 @@
+from sorcha.activity.base_activity import AbstractCometaryActivity
+import pandas as pd
+
+"""
+!!!!!!!!!!!!!!!!
+FOR TESTING ONLY
+!!!!!!!!!!!!!!!!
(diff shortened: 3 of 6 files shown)


pyproject.toml
```diff
@@ -105,7 +105,6 @@ write_to = "src/sorcha/_version.py"
 [tool.pytest.ini_options]
 testpaths = [
     "tests",
-    "src/sorcha",
```
drewoldag, line 108: Removed this since we no longer have any tests in the `src/sorcha` directory.

drewoldag:
@maxwest-uw - as it stands now, the `base_activity` and `activity_registration` modules are both very similar to the `base_lightcurve` and `lightcurve_registration` modules. Do you think it makes sense to try to abstract things up another layer and create a "AbstractModel" class that `AbstractLightcurve` and `AbstractCometaryActivity` inherit from. And a similar story around the registration modules - would it be worth parameterizing those so that I don't have the duplicate code? 

I'm on the fence, since I don't foresee any other plugin classes being developed so perhaps it's not worth it, but D.R.Y. alarms are also ringing in my head.

src/sorcha/activity/identity_activity.py
```diff
@@ -0,0 +1,55 @@
...
+        pd.DataFrame
+            The original ``observations`` dataframe, unchanged.
+        """
+
+        self._validate_column_names(df)
+
+        return df
+
+    @staticmethod
+    def name_id() -> str:
```
maxwest-uw, line 43: I think this could be changed to use the `@property` decorator

```suggestion
    @property
    def name_id() -> str:
```

that way we can access the `name_id` in a more pythonic way, like `activity.name_id`

src/sorcha/activity/base_activity.py
```diff
@@ -0,0 +1,44 @@
...
+        ----------
+        df : Pandas dataframe
+            The ``observations`` dataframe provided by ``Sorcha``.
+        """
+        for colname in self.required_column_names:
+            if colname not in df.columns:
+                raise (ValueError, f"Input dataframe is missing column %{colname}")
+
+    @staticmethod
+    @abstractmethod
```
maxwest-uw, line 41:
```suggestion
    @property
```

tests/activity/test_activity_registration.py
```diff
@@ -0,0 +1,42 @@
...
+
+    assert output == CA_METHODS
+
+
+def test_register_subclasses_with_duplicate():
+    class dup_identity(AbstractCometaryActivity):
+        def compute(self):
+            return 1
+
+        @staticmethod
```
maxwest-uw, line 21:
```suggestion
        @property
```

tests/activity/test_activity_registration.py
```diff
@@ -0,0 +1,42 @@
...
+
+    assert "duplicate cometary activity calculator name to CA_METHODS: identity" in str(excinfo.value)
+
+
+def test_update_subclasses():
+    class test_activity(AbstractCometaryActivity):
+        def compute(self):
+            return 1
+
+        @staticmethod
```
maxwest-uw, line 36:
```suggestion
        @property
```

mschwamb (approved):

src/sorcha/activity/identity_activity.py
maxwest-uw: aaaah I see, I assumed that we were instantiating the subclass, but if that's not the case then `@staticmethod` should work fine!

drewoldag: Cool. Thanks a bunch for your help, I really appreciate being about to talk through these ideas with you 😃

PR #527, opened 2023-08-18
Initial commit

### LINKED ISSUE #342: Things on the edges of CCDs won't be detected
Make the CCDs slightly smaller/check how far from the edge of the CCD - things that are less than 5 pixels from the edges of the CCD won't be detected - making the CCDs slightly smaller would deal with this or calculating the distances between the CCDs could do be the filter

jeremykubica:
Fixes #342 

This code adds a distance check for each object in a detector to the detector's closest edge. If the `edge_thresh` is provided and the distance is less than `edge_thresh`, the point is filtered.  Currently vectorizes over the points, but loops over the edges (there is likely a way to vectorize over both, but I'm keeping the first version simple).

Note: `edge_thresh` should be specified in the same units as the points (radians or degrees). Conversion from pixels will need to happen outside this function. This needs to be added to the configs in a later PR.


src/sorcha/modules/PPFootprintFilter.py
```diff
@@ -17,16 +17,53 @@
...
+
+    # Find the closest point on the line segment.
+    t = ((points[0] - x0) * (x1 - x0) + (points[1] - y0) * (y1 - y0)) / len_sq
+    t[t < 0.0] = 0.0
+    t[t > 1.0] = 1.0
+    proj_x = x0 + (x1 - x0) * t
+    proj_y = y0 + (y1 - y0) * t
+
+    # Compute the distances to the closest points on the line segment.
+    return np.sqrt((points[0] - proj_x) * (points[0] - proj_x) + (points[1] - proj_y) * (points[1] - proj_y))
```
drewoldag, line 64: This math makes me wish that docstrings could include drawings.

src/sorcha/modules/PPFootprintFilter.py
```diff
@@ -70,7 +107,7 @@ def __init__(self, points, ID=0, units="radians"):
         self.centerx = np.sum(self.x) / len(self.x)
         self.centery = np.sum(self.y) / len(self.y)
 
-    def ison(self, point, ϵ=10.0 ** (-11), plot=False):
+    def ison(self, point, ϵ=10.0 ** (-11), edge_thresh=None, plot=False):
```
drewoldag, line 110: I know that you didn't add it, but I wish that the original developer has used `epsilon`...

drewoldag (approved):
This looks pretty good to me. The test cases seem like they cover the edges (pun intended???). A question for @mschwamb - would it be valuable to include a message in the logs if some or all of the points are filtered due to being too close to the edges?

drewoldag:
Also, @jeremykubica if you haven't already created an issue to track the need for a config parameter for this, doing so would be very much appreciated!

mjuric:
Hi @mschwamb et al.:

I'm swamped right now & don't have time to review the math in detail, unfortunately :(. But go ahead and merge if this looks good to @drewoldag and @jeremykubica .

My high-level note (maybe for future upgrade): it looks like this implements point-in-polygon to test whether an object falls within the camera footprint? I'd advocate against implementing this manually -- use an existing (optimized) library. I did a bit of googling and Shapely seems promising (https://www.matecdev.com/posts/point-in-polygon.html, https://pypi.org/project/shapely/).

Alternatively, I have code from LSD that wraps a C implementation (it's fast): https://github.com/EUPSForge/polygon. Hasn't been updated since Python 2.7, but the math hasn't changed :). That said, think of this as a last resort; (much) better to use a strong 3rd-party library.

PR #549, opened 2023-08-23
Move the logging earlier

### LINKED ISSUE #544: Should logging start earlier?
If you screw up some command line arguments there is no change to the error log. Should logging start earlier?

jeremykubica:
Fixes #544 

Moves the logging earlier in the code flow to allow it to pick up info and errors during command line parsing. Also adds more comprehensive testing of the logging.

### DIFF
--- a/src/sorcha/modules/PPCommandLineParser.py
+++ b/src/sorcha/modules/PPCommandLineParser.py
@@ -110,4 +110,9 @@ def PPCommandLineParser(args):
             )
         )
 
+    # If verbose logging is on, log all the command line settings to INFO.
+    if args.v:
+        for flag, value in cmd_args_dict.items():
+            pplogger.info(f"Using commandline setting {flag} = {value}")
+
     return cmd_args_dict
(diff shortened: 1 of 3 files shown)


src/sorcha/modules/PPCommandLineParser.py
```diff
@@ -110,4 +110,9 @@ def PPCommandLineParser(args):
             )
         )
 
+    # If verbose logging is on, log all the command line settings to INFO.
```
mschwamb, line 113: It would be good if we change this so it's by default even if verbose logging is not on to ensure the user knows what was used with the output in the directory with their log file

mschwamb (changes requested):

src/sorcha/modules/PPCommandLineParser.py
jeremykubica: Sounds good. I made that change.

mschwamb (approved):

PR #553, opened 2023-08-24
Add an end to end diff test

### LINKED ISSUE #551: Create a test script with a set seed for people to check that their installation is correct
We want to have a script to run demo files and compare to expected output using a set seed, allowing the user to confirm that everything is working and can see within machine precision they're getting expected results.

jeremykubica:
Fixes #551

This PR replaces the integration test with and end to end diff test based on the demo data. Files are written into a temporary directory and compared with a new goldens result file. If the user makes a behavior impacting change, they need to manually run: `python src/sorcha/utilities/generateGoldens.py` to update the golden files (those updates will be part of their PR).

Since the diff test is written in the pytest framework, it will be part of the continuous integration.

WARNING: Numpy does not guarantee their random number generator is stable across versions of numpy, so goldens might need to be regenerated for different versions of numpy.


jeremykubica (commented):
One suggested change. << This was meant for another PR that I was reviewing in another tab.

src/sorcha/utilities/diffTestUtils.py
```diff
@@ -0,0 +1,74 @@
...
+        "surveyname": "LSST",
+        "outfilestem": f"out_end2end",
+        "verbose": False,
+    }
+    args = sorchaArguments(cmd_args_dict)
+
+    # Override the random number generator seed.
+    # WARNING: This is only accceptable in a test and should never be used for
+    # science results.
+    args._rng = np.random.default_rng(2023)
```
drewoldag, line 73: If I'm reading [this](https://numpy.org/neps/nep-0019-rng-policy.html#supporting-unit-tests) correctly, if we use `np.random.RandomState(2023)` instead of `default_rng(2023)`, then the random numbers should be consistent across numpy versions as well as operating platforms. 

From what I can tell, `RandomState` has been kept around specifically for this purpose with the understanding that it will not change between numpy version with the only exception being updates that are required to keep it working if numpy internals change.

drewoldag (approved):
This looks good to me. I left a comment about the use of `numpy.random.default_rng()` which should address the issue with consistency across numpy versions.

PR #555, opened 2023-08-25
Prefactor for the ASSIST+REBOUND Sprint

### LINKED ISSUE #495: Actually check the class of orbit from orbit file and return an error if it's not one of our support cases
We assume we're getting cometary or keplarian orbits into sorcha and don't check if it's not in the format we expected. So this did not crash https://github.com/dirac-institute/sorcha/blob/main/demo/tno_sample_1000_orbit.csv because the user added in a and e into the orbit file. This shouldn't work. We should error out because the orbit is carteisan. Error check on the format should happen in the orbit reader

### LINKED ISSUE #496: Support the read in of cartesian orbits
With the new ephemeris generator there should be no problem to include cartesian orbits, we don't actually cut on anything based on orbit, it was more not another thing to deal with and most people would draw from an orbital distribution.... though apparently not the DP0.3 folks ;-) - So we should get this fixed in the reader that it checks the full columns exist for the new ASSIST+REBOUND ephemeris generation

### LINKED ISSUE #497: improved orbit reader checks
- [ ] Check all orbit formats are the same for read in
- [ ] Check all the required elements for the specified orbit class exist 
- [ ] Remove any additional columns that are added that are beyond what we expect to read in
- [ ] Once #496 is implemented fix the benchmarks and demo files that have Cartesian and have an a and e tacked on to not crash the current reader

### LINKED ISSUE #515: Update line 119 in sorcha.py
Makes a reference to "comet", but should be `not None`

### LINKED ISSUE #545: Making the survey database be a command line argument
Since we're going to eventually be caching the pointing database if using the new ephemeris generator it makes sense to move the the pointing database as input on the command line. 

Also it is currently is the only input file controlled by the config file. We thought that a person will want to run the same pointing database multiple times, but what happens if you want to use this for a future survey and are running variations or running various versions of the rubin cadence simulations. 

Also this prevents someone from specifiying using the cache versus thinking they used the pointing database in the config file

maxwest-uw:
The results of me and @mschwamb 's pair programming session this wednesday. 

We made some changes to more strictly enforce orbit input types. You must now provide a format (cometary, keplerian, or cartesian) and pass along only the required parameters (defined in `inputs.rst`). Any other parameters may be passed in through the physical parameters file.

There is also a smattering of other changes that we got done, reflected in the list of issues below.

- closes #495 
- closes #496 
- closes #497 
- closes #515 
- closes #545 


src/sorcha/modules/PPCommandLineParser.py
```diff
@@ -31,8 +31,10 @@ def PPCommandLineParser(args):
     cmd_args_dict["configfile"] = PPFindFileOrExit(args.c, "-c, --config")
     cmd_args_dict["outpath"] = PPFindFileOrExit(args.u, "-u, --outfile")
 
-    if args.m:
-        cmd_args_dict["cometinput"] = PPFindFileOrExit(args.m, "-m, --comet")
+    if args.cp:
+        cmd_args_dict["complex_physical_parameters"] = PPFindFileOrExit(
+            args.cp, "-cp, --complex_phsyical_parameters"
```
jeremykubica, line 36: complex_phsyical_parameters -> complex_physical_parameters

jeremykubica (approved):

src/sorcha/readers/OrbitAuxReader.py
```diff
@@ -51,21 +52,46 @@ def _process_and_validate_input_table(self, input_table, **kwargs):
...
         super()._process_and_validate_input_table(input_table, **kwargs)
 
+        if len(input_table) == 0:
+            return input_table
+
         pplogger = logging.getLogger(__name__)
 
-        # Check for an H column (which should not be in the orbit file).
-        if "H" in input_table.columns:
+        if not "FORMAT" in input_table.columns:
```
drewoldag, line 60: Minor point - I find it a little easier to read as: `if "FORMAT" not in input_table.columns:`

src/sorcha/readers/OrbitAuxReader.py
```diff
@@ -51,21 +52,46 @@ def _process_and_validate_input_table(self, input_table, **kwargs):
...
+            pplogger.error("ERROR: PPReadOrbitFile: Orbit format must be provided.")
+            sys.exit("ERROR: PPReadOrbitFile: Orbit format must be provided.")
+
+        if len(input_table.columns) != 9:
+            pplogger.error("ERROR: Please only provide the required columns in your orbits file.")
+            sys.exit("ERROR: Please only provide the required columns in your orbits file.")
+
+        orb_format = input_table["FORMAT"].iloc[0]
+        if not np.all(orb_format == input_table["FORMAT"].values):
```
drewoldag, line 69: Also minor point - This is conditional is probably fast enough (depending on how many times it's called), but pandas provides a `unique` method that may be more efficient.

`if len(input_table["FORMAT"].unique()) != 1:`

A reference SO response related to using `set`s to find unique values: 
https://stackoverflow.com/a/51227605

drewoldag (approved):
This looks like a nice refactor to me! Thanks a bunch for getting all these issues resolved!

PR #560, opened 2023-08-29
A series of improvements to the argument parsing

### LINKED ISSUE #557: adjust argpase initalization so it outputs defaults for variables when --help is used
Currently the command line help looks like this:

Screenshot 2023-08-29 at 3 27 10 AM

It's not clear what the defaults are for some of these variables. Like the survey option. The default is LSST so it doesn't ever need to be set right now, but I would not know from looking at the help output. 

It looks like there is an easy way to update this so the help string outputs the default value. 

https://stackoverflow.com/questions/12151306/argparse-way-to-include-default-values-in-help

### LINKED ISSUE #558: Separate out the required command line variables when --help prints out
There's a lot of config file variables and it's not clear what's actually required to make things go. 
These are then ones at minimum that will be needed  So it would be good if the --help on the command line could state these are the required ones  in a separate section from the rest of the command line configuration options
  
 -c C, --config C      Input configuration file name
  -p P, --params P      Physical parameters file name
  -ob OB, --orbit OB    Orbit file name
  -pd PD, --pointing_database PD
  -o O, --outfile O     Path to store output and logs.

Looks like that's doable according to stack overflow https://stackoverflow.com/questions/24180527/argparse-required-arguments-listed-under-optional-arguments

jeremykubica:
Fixes #557
Fixes #558

A series of small improvements and fixes to the command line argument parsing.
1) Fixes conflicting merges between #555 and some other PR (such as #553). In particular #555 changed the name of some of the command line arguments (`o` -> `ob` and `u` -> `o`). This updates them in the additions from the other PRs.
2) Print out the default values for --help (fixes #557)
3) Separate out the required vs. optional parameters for --help (fixes #558)
4) Add pointing database argument to `PPCommandLineParser`
5) List arguments in alphabetical order within optional/required sections.

Describe your changes.

### DIFF
--- a/src/sorcha/sorcha.py
+++ b/src/sorcha/sorcha.py
@@ -371,12 +371,10 @@ def main():
         dest="cp",
     )
     optional.add_argument(
-        "-dw",
-        help="Make temporary ephemeris database. If no filepath/name supplied, default name and ephemeris input location used.",
-        dest="dw",
-        nargs="?",
-        const="default",
-        type=str,
+        "-dl",
+        help="Delete the temporary ephemeris database after code has completed.",
+        action="store_true",
+        default=False,
     )
     optional.add_argument(
         "-dr",
@@ -385,10 +383,12 @@ def main():
         type=str,
     )
     optional.add_argument(
-        "-dl",
-        help="Delete the temporary ephemeris database after code has completed.",
-        action="store_true",
-        default=False,
+        "-dw",
+        help="Make temporary ephemeris database. If no filepath/name supplied, default name and ephemeris input location used.",
+        dest="dw",
+        nargs="?",
+        const="default",
+        type=str,
     )
     optional.add_argument(
         "-f",
--- a/src/sorcha/sorcha.py
+++ b/src/sorcha/sorcha.py
@@ -302,8 +302,9 @@ def main():
           -f, --force        Force deletion/overwrite of existing output file(s). Default False.
     """
 
-    parser = argparse.ArgumentParser()
-    parser.add_argument(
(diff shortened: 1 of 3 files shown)


tests/sorcha/test_PPCommandLineParser.py
```diff
@@ -7,10 +7,11 @@
 class args:
     def __init__(self, cp, dw, dr, dl, t="testout", u="./", f=False):
         self.p = get_test_filepath("testcolour.txt")
-        self.o = get_test_filepath("testorb.des")
+        self.ob = get_test_filepath("testorb.des")
         self.e = get_test_filepath("oiftestoutput.txt")
         self.c = get_test_filepath("test_PPConfig.ini")
-        self.u = u
+        self.pd = get_test_filepath("baseline_10klines_2.0.db")
+        self.o = u
```
drewoldag, line 14: How much more work would it be to change the input parameter name from `u` to `o` here?

src/sorcha/sorcha.py
```diff
@@ -372,38 +362,59 @@ def main():
...
+        dest="v",
+        default=True,
+        action="store_false",
+    )
+
     args = parser.parse_args()
 
     # Extract the output file path now in order to set up logging.
-    outpath = PPFindFileOrExit(args.u, "-u, --outfile")
+    outpath = PPFindFileOrExit(args.o, "-o, --outfile")
```
drewoldag, line 417: Thanks for cleaning this up.

drewoldag (approved):
Looks good to me. There are a ton of input args, so this should really help future users.

tests/sorcha/test_PPCommandLineParser.py
jeremykubica: Good call. I fixed throughout the test.

Issue #466, opened 2023-07-25
Improve the linking filter - if that fails create quick and dirty SSP detection efficiency filter


mschwamb:
Difi is slow but for most populations if there are enough chances you don't need to do it that thoroughly. So if there are enough observations for linking to happen at least once and over N years than we can say all the precovery would be done and find all detections on Rubin images. Then we would need to remove X% of objects based on the linking efficiency of SSP. 

How about @bernardinelli, @astronomerritt, and I sketch this out?

mschwamb:
I think this is more complicated than what you wrote @astronomerritt .

This is on objects not observations now, but with some quick checks/cuts on the number/distribution of detections.    What happens if some object only has 2 tracklets and never got a third within 14 days? that wouldn't have been linked, but I think the above says it would be counted as a discovery if random_chance < linking_efficiency

astronomerritt:
Ah, so you still want to keep _some_ of the linking filter checks in a sense?

I wonder if it can be done statistically by number of observations of the object. For example if we could determine that if an object is observed ~100 times it's 95% per cent likely to be linked but if it's observed 5 times that's 0%. Some testing might help us come up with an empirical function.

Alternatively we could check for at least six observations in 14 days, or more rigorously at least three tracklets in 14 days, but once we start introducing that rolling window the runtime tends to go up - at least it does when I do it!

mschwamb:
Currently @mjuric  believe's that it's better to see if we can implement a full linking filter instead. So he's going to work on improving difi or adapting his version for linking for here.

PR #566, opened 2023-08-31
Switch PPLinkingFilter to "mini-difi"

### LINKED ISSUE #573: Set random seed secretly by environment variable if the environment variable is defined.
Based on feedback it would be easier if the random seed could be set by secret environment variable

mjuric:
closes #573, #466

Implements an algorithm for mock linking that should be faster than what we have at the time. Synthetic tests indicate that this should be able to link a full S3M 10yr LSST sim on about ~20 minutes on an M1 Max core (this counts only the cost of linking, not the rest of the `sorcha` code).

The long-term solution is still to speed up the implementation in `difi`; consider this a stopgap.

### DIFF
--- a/src/sorcha/modules/PPLinkingFilter.py
+++ b/src/sorcha/modules/PPLinkingFilter.py
@@ -96,6 +96,7 @@ def PPLinkingFilter_OLD(
 
     return linked_object_observations
 
+
 def PPLinkingFilter(
     observations,
     detection_efficiency,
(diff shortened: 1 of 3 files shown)


mjuric:
@mschwamb this is ready to be reviewed. Note I intentionally left the old code in the file (renamed the function to `PPLinkingFilter_OLD`) so I can more easily make some A/B tests and benchmarks while we're still actively developing.

mschwamb:
Awesome! I added the tickets that this PR closes as well.

drewoldag (approved):
There's a lot here, and I don't have enough context to know if miniDifi is being thoroughly tested. But since the old version of the code is in place, I'm assuming that baseline vs. new has been compared and are producing identical results.

mjuric:
It passes the same unit tests, but you're right that thorough testing is the next thing on the TODO list.

Thanks for the merge!

PR #577, opened 2023-09-01
Issue #509: light curve models use complex parameters input file.

### LINKED ISSUE #509: Have light curve parameters inputted through the comet parameters file
Writing this, so I don't forget for the Wednesday meeting:

A philosophical question about inputs - currently we handle light curves and cometary activity differently in terms of where things are inputted. 

The user specifies the light curve columns in the physical parameters inputted For cometary activity, we have a separate file inputted from the command line. The reason for this was that is very possible that the user wants to run different cometary activity values against the same population of objects and see how things change. This way the user only needs to worrya bout swapping out the file pointing to the relevant cometary columns rather than editing the lack two columns of a large physical parameters csv file. 

The comet people we talked to that thought that was a good idea. Also it meant that the documentation was the same for the physical parameters file. It would always be in one format no matter the configuration you're running in (activity or no activity).

So should the comet parameters input file still exist? Should we call it instead the advanced_parameters_file and it this is where the light curve parameters should also go? I think yes, but it would be good to get everyone experts opinions

astronomerritt:
Fixes #509.

Comet activity and light curve models now use the same file for input parameters: the complex parameters input file. Facilitating this required one very small change to sorcha.py: now when lc_model is not None, the code looks for the complex parameters input file and merges it onto the dataframe.

### DIFF
--- a/src/sorcha/sorcha.py
+++ b/src/sorcha/sorcha.py
@@ -132,7 +132,7 @@ def runLSSTPostProcessing(cmd_args, pplogger=None):
 
     reader.add_aux_data_reader(OrbitAuxReader(args.orbinfile, configs["aux_format"]))
     reader.add_aux_data_reader(CSVDataReader(args.paramsinput, configs["aux_format"]))
-    if configs["comet_activity"] is not None:
+    if configs["comet_activity"] is not None or configs["lc_model"] is not None:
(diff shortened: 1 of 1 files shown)


drewoldag (approved):
Looks good. Thanks!

PR #581, opened 2023-09-01
some more rng changes

### LINKED ISSUE #563: remove the random seed variable being read in from the config file
@maxwest-uw set up the random seed to be set not from the config file, but it looks like we're still reading it in if in the config file (though I think the variable is not being used). I think these lines need to be removed

https://github.com/dirac-institute/sorcha/blob/a707f524d1eaac51da28ae684993baf5592c941b/src/sorcha/modules/PPConfigParser.py#L672C1-L672C20

### LINKED ISSUE #567: random seed needs to be outputted to the log file
This might have been removed in the refactoring.

 https://github.com/dirac-institute/sorcha/blob/a1cdab27b7a9b985e2573566a1f3bea5709e013b/src/sorcha/utilities/sorchaArguments.py#L40C1-L40C51
The seed for the random number generator needs to be outputted to prominently to the log file so that if the user wants to rerun at a later date and get the same answer they can.

maxwest-uw:
Fixes #563 , fixes #567 .

Removes the rng_seed read in in `PPConfigParser`, creates a new hidden-ish variable in `sorchaArguments` to keep track of the rng seed, and prints out the seed in `pplogger`.

Note that this slightly changes the API of manually setting the rng_seed. You can still set the rng generator to use a seed of your choosing but if you don't also change the `_rng_seed` variable it will print out the default seed based on time, i.e. not the one that's actually being used in the numpy rng. (we could add a method to handle both changes at once and "refresh" the rng with a new seed but adding that functionality seems to go against the purpose of hiding it in the first place, right?)

### DIFF
--- a/src/sorcha/modules/PPConfigParser.py
+++ b/src/sorcha/modules/PPConfigParser.py
@@ -668,13 +668,6 @@ def PPConfigFileParser(configfile, survey_name):
         config, "EXPERT", "pointing_sql_query", "ERROR: no pointing database SQLite3 query provided."
     )
 
-    if config.has_option("EXPERT", "rng_seed"):
-        config_dict["rng_seed"] = PPGetIntOrExit(
-            config, "EXPERT", "rng_seed", "ERROR: this error should not trigger."
-        )
-    else:
-        config_dict["rng_seed"] = None
-
     config_dict["lc_model"] = config.get("EXPERT", "lc_model", fallback=None)
     config_dict["lc_model"] = None if config_dict["lc_model"] == "none" else config_dict["lc_model"]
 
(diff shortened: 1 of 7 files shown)


mschwamb:
#566 does some refactoring of the random seed. So this should wait until that goes in and that rebase. I am not sure there will be much needed to be updated here, but worth doing things in that order I think

mschwamb:
Looks like #582 needs to be merged first

mschwamb (changes requested):
Conflicts need to be addressed first

maxwest-uw:
there were multiple reworks of how rng works after this pr was opened so I'm just going to close this and redo the changes since that's easier than merging 😅

PR #592, opened 2023-09-06
Overhaul the random number generation

### LINKED ISSUE #585: overhaul random number generation
From @mjuric  

While messing with the RNG, I noticed one (design) improvement we should think about. The problem: Our RNG is globally shared, between all modules/parts of the code. Therefore, turning on/off one module, or changing its implementation, changes the RNG state for all other modules that run after it. This means that a module may indirectly alter the output of completely unrelated modules.

Example: say we have magnitude error, astrometric error, and the linking module. Turning off magnitude error changes the state of the RNG that is used for the astrometric error which then generates slightly different astrometry, which affects linking results.

Solution: each module should have its own RNG, and initialize it using the global seed + some constant unique to the module (maybe a hash of its name). That way plugging modules in/out doesn't cause "spooky RNG action at a distance" in a different module.

Why do we care: a) it's easier to test/validate the results, and b) simulations can be rerun with additional modules (or fixed modules) in a way that doesn't change the numbers in unrelated columns.

jeremykubica:
Fixes #585 

Changes the random number generation to use per-module random number generators instead of a single global generator. The goal is to provide consistent random numbers for testing regardless of which modules are turned off (see ticket #585 for details).

Creates a class `PerModuleRNG` that takes a base seed and stores per-module random number generators. Each RNG is seeded based on the base seed and the module name.

Describe your changes.


src/sorcha/modules/PPModuleRNG.py
```diff
@@ -0,0 +1,38 @@
...
+        self._rngs = {}
+
+    def getModuleRNG(self, module_name):
+        """
+        Return a random number generator that is based on a base seed
+        and the current module name.
+
+        Parameters:
+        -----------
+        base_seed (int): The base seed for a random number generator
```
drewoldag, line 24: This should be `module_name`.

drewoldag (approved):
This is really cool I like this utility a lot! I left one comment about a docstring, but otherwise this looks fine. 

Also, I think this could make for a neat lightning talk or something for our Thursday meetings. It seems like the type of thing that is applicable to lots of our projects at this point.

src/sorcha/modules/PPModuleRNG.py
jeremykubica: Fixed

Issue #596, opened 2023-09-07
Merge runLSSTPostProcessing and runLSSTSimulation 


mschwamb:
Adding a formal ticket for merging runLSSTPostProcessing and runLSSTSimulation  and do the logic of whether ingesting a CSV file for ephemeris or using A+R in the combined function

mschwamb:
#559 ated

drewoldag:
This of items for this issue:

- [ ] Make use of the `-f` flag and warn the user if they are about to overwrite the ephemeris file
- [ ] Add new CLI argument for ephemeris output file name
- [ ] Add logic in main sorcha `while` loop to either call A&R for the chunk or read the ephemeris file chunk
- [ ] Return pandas dataframe from A&R. Append output of A&R to output file (due to chunking, don't want to overwrite the output for each chunk.)
- [ ] Provide a CLI argument to define where the A&D data files were downloaded.

mschwamb:
Note because of possible chunking the  output the ephemeris csv file needs to be  appending to the file so that if the process is in the second or third iteration of the loop that it keeps adding to the bottom

PR #603, opened 2023-09-08
WIP - A&R accepts chunked input, returns dataframe

drewoldag:
A&R accepts chunked dataframe as input, returns observation dataframe, also outputs .csv file.

Addresses #596 

Removed the separate function in sorcha.py that was calling A&R. That call to A&R is now buried inside the `while` loop that chunks the input file. 

If the user specifies "ar" as the ephemeris type in the config file, A&R will be run on the input, and return a pandas dataframe as well as output a CSV file. If the user specifies "external" as the ephemeris type, sorcha will look for a previously generated ephemeris file, and chunk on that without calling A&R.


src/sorcha/readers/CombinedDataReader.py
```diff
@@ -128,10 +128,78 @@ def read_block(self, block_size=None, verbose=False, **kwargs):
...
-                sys.exit("ERROR: At least one missing ObjID {reader.get_reader_info()}")
+                pplogger.error(f"ERROR: At least one missing ObjID in {reader.get_reader_info()}")
+                sys.exit(f"ERROR: At least one missing ObjID {reader.get_reader_info()}")
 
             verboselog("Joining auxiliary data with ephemeris")
             ephem_df = ephem_df.join(current_df.set_index("ObjID"), on="ObjID")
 
         return ephem_df
+
+    def read_aux_block(self, block_size=None, verbose=False, **kwargs):
```
drewoldag, line 139: I don't particularly like this function - it seems like we should be able to use `read_block` instead of creating a new, almost identical version. @jeremykubica I would like to discuss this with you when you have a few minutes :)

src/sorcha/sorcha.py
```diff
@@ -189,7 +162,12 @@ def runLSSTPostProcessing(cmd_args, configs, pplogger=None):
         verboselog("Working on objects {}-{}.".format(startChunk, endChunk))
 
         # Processing begins, all processing is done for chunks
-        observations = reader.read_block(block_size=configs["size_serial_chunk"])
+        if configs["ephemerides_type"].casefold() == "external":
```
drewoldag, line 165: This is the logic that will either read from an external ephemeris file, or pass in a chunk of the orbits dataframe for A&R to process.

drewoldag:
The current issue is that the end to end test fails to merge the ephemeris dataframe and the input orbits dataframe because ObjID for ephemeris is type "int64" and ObjID for the input orbits dataframe is type "object". I'll look into that next week.

src/sorcha/ephemeris/simulation_driver.py
```diff
@@ -137,11 +160,57 @@ def create_ephemeris(args, configs):
...
+                        dobs_sundt[0] * AU_KM / (24 * 60 * 60),
+                        dobs_sundt[1] * AU_KM / (24 * 60 * 60),
+                        dobs_sundt[2] * AU_KM / (24 * 60 * 60),
+                        phase_angle * 180 / np.pi,
+                    )
+
+                    in_memory_csv.writerow(out_tuple)
+
+    # reset to the beginning of the in-memory CSV
+    output.seek(0)
```
maxwest-uw, line 192: when I initially saw this I assumed that writing in memory and then reading the in mem csv to a dataframe would be slower than just adding a row to an existing dataframe, but I ran the numbers and it's actually faster by a factor of ~27 🤯

src/sorcha/ephemeris/simulation_setup.py
maxwest-uw, line 1: there's also a `make_retriever` in `furnish_spiceypy` that needs the `args.ar_data_file_path`

src/sorcha/ephemeris/simulation_driver.py
drewoldag: Yes, my understanding here is that any time you append to a dataframe a new copy of the data structure has to be made, so with small dataframes it's not a big deal, but as the size increases it becomes more efficient to append to a csv, then convert to a df.

maxwest-uw (approved):

demo/PPConfig_test.ini
```diff
@@ -5,7 +5,7 @@
 
 # The simulation used for the ephemeris input.
 # Options: "ar", "external"
-ephemerides_type = ar
+ephemerides_type = external
```
drewoldag, line 8: @mschwamb This is one of the config files that is rendered in the documentation. As you can see, I've changed the ephem_type here. Please let me know if this is a problem. 

Part of the reason for doing this was that the end to end test was failing when ephem_type = 'ar'. (Which makes sense, because the e2e test was created before A&R was built in)

mschwamb: That's fine. I'll update the documentation to use sorcha_demo.ini instead. Sorry, I thought the end to end test is running in CI and was using a different config file

src/sorcha/sorcha.py
```diff
@@ -43,36 +43,6 @@
...
-        args.validate_arguments()
-    except Exception as err:
-        pplogger.error(err)
-        sys.exit(err)
-
-    configs = PPConfigFileParser(args.configfile, args.surveyname)
-    create_ephemeris(args, configs)
-
-
 def runLSSTPostProcessing(args, configs, pplogger=None):
```
mschwamb, line 46: This function will no longer be only postprocessing, so I think it would be good to change runLSSTPostProcessing to  runLSSTSimulation in terms of name, if possible.

drewoldag: Done!

mschwamb (approved):

PR #604, opened 2023-09-08
add rng seed printing and remove config options

### LINKED ISSUE #563: remove the random seed variable being read in from the config file

### LINKED ISSUE #567: random seed needs to be outputted to the log file

maxwest-uw:
Fixes #563, fixes #567 .

Describe your changes.

### DIFF
--- a/src/sorcha/modules/PPConfigParser.py
+++ b/src/sorcha/modules/PPConfigParser.py
@@ -699,13 +699,6 @@ def PPConfigFileParser(configfile, survey_name):
         config, "EXPERT", "pointing_sql_query", "ERROR: no pointing database SQLite3 query provided."
     )
 
-    if config.has_option("EXPERT", "rng_seed"):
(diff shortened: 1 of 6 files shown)


drewoldag (approved):
Looks good

PR #632, opened 2023-09-14
Pre-computing all of the pointing database-related calculations for A&R.

### LINKED ISSUE #622: Store in memory the pointing information calculation for ephemeris generation if sorcha is reading in chunks of input
To me,  [theses lines](https://github.com/dirac-institute/sorcha/blob/305b23d486d2352491cbd306afa067ad45c33c47/src/sorcha/ephemeris/simulation_driver.py#L50C6-L70C1) are what is computed for the pointings and the Sun's location based on this. 

If sorcha.py is going through the orbit/physical parameters file in chunks then those values above could be returned from create_ephemeris, and if inputted into the function that means we don't need to recompute these values

drewoldag:
Fixes #622 .

Pulled most of the per-pointing calculations out of the main for loop in A&R into a pre-compute function that is run once. The pre-calculated values are stored on the pointing dataframe, and passed to A&R for each chunk of data to be processed.

If the chunk size is large (and data size is small) such that only one chunking of data is required, there will be no performance improvement. As a benchmark though, using the example 1 year pointing database on my local machine, the pre-computation time was 12-13 seconds. I think this can be extrapolated to ~120-130 seconds for a 10 year database. Meaning that we would save approximate 2 minutes per chunk of data processed (for chunk number >1).

### DIFF
--- a/src/sorcha/ephemeris/simulation_driver.py
+++ b/src/sorcha/ephemeris/simulation_driver.py
@@ -7,7 +7,11 @@ import numpy as np
 import pandas as pd
 import spiceypy as spice
 
-from sorcha.ephemeris.simulation_setup import *
+from sorcha.ephemeris.simulation_setup import (
+    create_assist_ephemeris,
+    furnish_spiceypy,
+    generate_simulations,
+)
 from sorcha.ephemeris.simulation_constants import *
 from sorcha.ephemeris.simulation_geometry import *
 from sorcha.ephemeris.simulation_parsing import *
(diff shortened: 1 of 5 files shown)


src/sorcha/ephemeris/simulation_driver.py
```diff
@@ -234,26 +199,3 @@ def update_pixel_dict(jd_tdb, t_picket, picket_interval, sim_dict, ephem, obsCod
         this_pix = hp.vec2pix(nside, rho_hat[0], rho_hat[1], rho_hat[2], nest=True)
         pixel_dict[this_pix].append(k)
     return t_picket, pixel_dict, r_obs
-
-
-def barycentricObservatoryRates(
```
drewoldag, line 239: Moved to `simulation_geometry` to avoid circular dependency.

src/sorcha/ephemeris/simulation_driver.py
```diff
@@ -70,98 +74,59 @@ def create_ephemeris(orbits_df, pointings_df, args, configs):
     for _, pointing in pointings_df.iterrows():
         mjd_tai = float(pointing["observationStartMJD"])
         ra, dec = float(pointing["fieldRA"]), float(pointing["fieldDec"])
-        visit_vec = ra_dec2vec(ra, dec)
```
drewoldag, line 73: This is the large block of code that was called out in issue #622 as something that could be pre-computed.

src/sorcha/ephemeris/simulation_driver.py
```diff
@@ -70,98 +74,59 @@ def create_ephemeris(orbits_df, pointings_df, args, configs):
...
+                    r_ast_sun = r_ast - pointing["r_sun"]
+                    v_ast_sun = v_ast - pointing["v_sun"]
+                    r_ast_obs = r_ast - pointing["r_obs"]
                     phase_angle = np.arccos(
                         np.dot(r_ast_sun, r_ast_obs) / (np.linalg.norm(r_ast_sun) * np.linalg.norm(r_ast_obs))
                     )
-                    obs_sun = r_obs - r_sun
-                    dobs_sundt = v_obs - v_sun
-
-                    outstring = "%s,%s," % (obj_id, pointing["FieldID"])
```
drewoldag, line 138: `outstring` code was not being used.

src/sorcha/ephemeris/simulation_driver.py
```diff
@@ -70,98 +74,59 @@ def create_ephemeris(orbits_df, pointings_df, args, configs):
...
-            sun = ephem.get_particle("Sun", jd_tdb - ephem.jd_ref)  # This can be pre-calculated
-            r_sun = np.array((sun.x, sun.y, sun.z))
-            v_sun = np.array((sun.vx, sun.vy, sun.vz))
 
         # If the observation time is too far from the
         # time of the last set of ballpark sky position,
         # compute a new set
         while (
-            abs(jd_tdb - t_picket) > 0.5 * picket_interval or first == 1
+            abs(pointing["jd_tdb"] - t_picket) > 0.5 * picket_interval or first == 1
```
drewoldag, line 82: Most of the changes from here down are simply making use of the precomputed values stored in the `pointing` dataframe row.

matthewholman:
A savings of 120 sec for a 10 year simulation is consistent with what I measured.

src/sorcha/ephemeris/simulation_setup.py
```diff
@@ -104,3 +116,80 @@ def generate_simulations(ephem, gm_sun, orbits_df):
         sim_dict[row["ObjID"]]["ex"] = ex
 
     return sim_dict
+
+
+def precompute_pointing_information(pointings_df, args, configs):
```
drewoldag, line 121: This does the precomputation work. There are two instances of `pandas.apply` being used in here. Both of them are a result of not being able to _easily_ vectorize the code. 

However, it should be noted that with some additional time, I do think that we could vectorize those two calls and get a little more speed out of this.

src/sorcha/sorcha.py
```diff
@@ -165,6 +166,10 @@ def runLSSTSimulation(args, configs, pplogger=None):
             observations = reader.read_block(block_size=configs["size_serial_chunk"])
         else:
             orbits_df = reader.read_aux_block(block_size=configs["size_serial_chunk"])
+            # determine if the pointing dataframe has been primed for creating ephemeris
+            if "visit_vector" not in filterpointing:
```
drewoldag, line 170: Please let me know if there's a better way to guard this. Right now the code just checks for one of the columns that would have been added if the pre-computation step had already happened.

matthewholman: I have generally used dictionaries for this kind of thing.  The way I have kept track of visits with multiple telescopes is to use a key that is a tuple of the observatory code and the date/time of the exposure.   That should always be unique and corresponds one-to-one with a set of the precalculations you are doing.  Not sure if that helps you, but maybe you can use that idea to help keep track of what has already been done.

src/sorcha/ephemeris/simulation_setup.py
jeremykubica: In addition, there are a few `for` loops over the rows (in the form of `enumerate`), but I don't think it is worth continuing to optimize them unless they turn into a bottleneck.

src/sorcha/sorcha.py
jeremykubica: Why not just do this precomputation when to first create `filterpointing` (Around line 118)

You could still use the `configs["ephemerides_type"]` check from line 164.

jeremykubica (approved):

maxwest-uw (approved):

src/sorcha/sorcha.py
drewoldag: Ooohhh yeahhhh. Yes, that's a great point @jeremykubica thanks for the suggestion!

PR #644, opened 2023-09-21
Adding ephemeris generation unit test

### LINKED ISSUE #640: unit test for ephemeris generation
for the 1 year baseline database in demo 

[sorcha_ephemeris.csv](https://github.com/dirac-institute/sorcha/files/12666148/sorcha_ephemeris.csv)
[params_small_random_mpcorb.csv](https://github.com/dirac-institute/sorcha/files/12666149/params_small_random_mpcorb.csv)
[orbits_small_random_mpcorb.csv](https://github.com/dirac-institute/sorcha/files/12666150/orbits_small_random_mpcorb.csv)

using the values in demo/sorcha_demo_config.ini

drewoldag:
Fixes #640 .

Using an approach similar to what was established for the end to end unit test for sorcha. We'll output the results of ephemeris generation and then compare it against a "golden output" file. 

I needed to update the script that generates the golden output as well, so you'll see changes in that utility module. 

Additionally, there was a function that compared the test vs. the golden outputs. It was originally written to read in the .csv outputs of the test and golden data as numpy arrays and then compare them with specific logic to compare numeric and string types differently. However, due to the way that the test and golden data was ingested by numpy, all values were assigned a `str` type, so an absolute comparison was being performed.

Thus I modified this comparison function to read the files into pandas dataframes and used the `pandas.api.types.is_numeric_dtype` function to deteremine columns should be compared absolutely (as strings) or using numpy.allclose (for numeric types).


demo/sorcha_config_demo.ini
mschwamb, line 1: So there's been a lot of various version of config file versions running, let's make this it's own config file. My reason for this is that sorcha_config_demo.ini having ar_ang_fov = 1.8 might confuse people as the correct value for LSST needs to be 2.06. Matt and Steph validated the 1.8 fov so my suggestion is my a goldens config file and leave sorcha_config_demo.ini alone/be separate for tutorials

jeremykubica (approved):

src/sorcha/utilities/diffTestUtils.py
```diff
@@ -54,24 +82,19 @@ def override_seed_and_run(outpath):
...
-        "pointing_database": get_demo_filepath("baseline_v2.0_1yr.db"),
-        "outpath": outpath,
-        "makeTemporaryEphemerisDatabase": False,
-        "readTemporaryEphemerisDatabase": False,
-        "deleteTemporaryEphemerisDatabase": False,
-        "surveyname": "LSST",
-        "outfilestem": f"out_end2end",
-        "verbose": False,
-    }
+    if arg_set is "baseline":
```
jeremykubica, line 85: Should this be ==?

demo/sorcha_config_demo.ini
jeremykubica: Agreed.

src/sorcha/utilities/diffTestUtils.py
```diff
@@ -54,24 +82,19 @@ def override_seed_and_run(outpath):
...
+        cmd_args_dict = BASELINE_ARGS
+    elif arg_set is "with_ephemeris":
```
jeremykubica, line 87: Same with ==

src/sorcha/utilities/diffTestUtils.py
drewoldag: Yep, fixed!

src/sorcha/utilities/diffTestUtils.py
drewoldag: Also yep, also fixed :)

demo/sorcha_config_demo.ini
drewoldag: Sure, I can do this.

PR #645, opened 2023-09-21
make footprint_path config optional

### LINKED ISSUE #541: Rename detectors_corners.csv to something clearer LSST_detector_corners.csv and allow users to use the version that's pip/conda installed by default
Giving a more specific name to the LSST_detector_corners_082023.csv and how we set where the corners file path  so that it matches what is done for the files downloaded by pooch. 

Since we now you can pip install with downloading the source code, sorcha should read in that pip installed/conda installed  corners file by default unless a user specifies a different file in the config. I am not sure what the config file logic should be here so that it uses what's available or otherwise the one in the config file.

maxwest-uw:
Keeps the functionality of allowing a user to manually pass their own footprint file, but also make that optional and pull the file from our own local copy.

Fixes #541  

Note: I had to move one of the copies of `detectors_corners.csv` into a new `modules/data/` directory to make everything work nicer with `pkg_resources`. Theoretically we could've manually determined the path of the version in the `/demo/` directory but that would require a lot of messy path manipulation and I'm not sure if it would work with the pip/conda version of `sorcha`. `pkg_resources` is the recommended way to refer to data files within your package and for some reason they don't allow you to refer to files outside of the module you're working in 🙃.


drewoldag (approved):
Sweet, looks good to me.

mschwamb:
Is it possible to address the other part of this ticket as well to rename the csv file to something more specific since detectors_corners.csv is very generic? Renaming the file to  LSST_detector_corners_100123.csv?

mschwamb:
Two other minor things if possible to consider/address-

The LSSTCam footprint is only valid for the LSST survey. It won't be for any other surveys. I know the code isn't set up for other surveys really, but the footprint filter actually is. So this one function someone else could pull because it's generic.  it would be good that we specific here if possible for futureproofing.  The[ cmd line argument survey is set.](https://github.com/dirac-institute/sorcha/blob/65e6362d9441128810ef211b5e52710a005286cb/src/sorcha/sorcha.py#L477C29-L477C29) It would be good that only if it is set to LSST/lsst is the internal file to  be used or there's an error thrown because there is no internal file for any other surveys. 

If I read this correctly, as a user if I look at my config file I'd remove footprint_path from my config file to use the internal. My worry is there's no easy way a user who hasn't read the documentation or hasn't read it a long while knows that they're using the internal file. Would it be clearer to setup the samething for ephemeris? A config file that says where using external or internal?

maxwest-uw:
went ahead and changed the logic to check for the survey name when loading the default and changed the name of the detectors file to be more specific, but I'm kinda opposed to adding another enabled/disabled variable especially for a single small optional variable, I think as long as we say in the docs that the `footprint_path` is optional for LSST then I think that's probably fine (the worst case scenario is that a current user of `sorcha` continues using their local copy of the detectors file, which doesn't really affect how the package runs). possible compromise could be to make the user use a keyword like `footprint_path = default` instead

drewoldag:
This PR still looks good. It might be a wise idea to rebase from `main` prior to squash-n-merge-ing to make sure that the variable renaming work doesn't accidentally clobber anything here.

PR #649, opened 2023-09-26
Update installation.rst to pin rebound for conda create

mschwamb:
Pinning the version of rebound at least in the conda create command to help with m1/m2 macs

Fixes # .

Describe your changes.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -42,7 +42,7 @@ Setup Your Conda Environment
 
 **Step 1** Create a conda environment::
 
-   conda create -n sorcha -c conda-forge -c moeyensj numpy numba pandas=2.1 scipy astropy matplotlib sbpy pytables difi==1.2rc3 spiceypy healpy rebound pooch tqdm h5py python=3.10
+   conda create -n sorcha -c conda-forge -c moeyensj numpy numba pandas=2.1 scipy astropy matplotlib sbpy pytables difi==1.2rc3 spiceypy healpy rebound=3.26.3 pooch tqdm h5py python=3.10
 
 .. tip::
    We recommend using python version 3.9 or higher with Sorcha. The conda command uses python 3.10.


mjuric:
There's a conflict between the conda environment specs & the python packaging specs. pyproject.toml specifies pandas<2.1 and pins a specific numpy. Should that be changed to:

```diff
(sorcha) [mjuric@caladan (main) ~/projects/github.com/dirac-institute/sorcha]$ git diff
diff --git a/pyproject.toml b/pyproject.toml
index d36dc24..9380317 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,14 +21,14 @@ classifiers = [
 ]
 dynamic = ["version"]
 dependencies = [
-    "numpy==1.24.3",
-    "pandas<2.1",
+    "numpy>=1.24.3",
+    "pandas>=2.1",
     "scipy",
     "astropy",
     "matplotlib",
     "sbpy",
     "tables",
-    "difi == 1.2rc3",
+#    "difi == 1.2rc3",
     "spiceypy",
     "healpy",
     "assist",
```
?

mjuric:
Looking elsewhere in the docs, I'd propose to let the code store the cache files into the default directory, and not override it (right now it goes to `./ar_files`).

Default should be sane things that the users will use; we shouldn't be training them to override the defaults w/o a good reason. (it's also may make them think that they /have/ to override them). A third benefit is that it simplifies the instructions.

Propose to change that too.

mschwamb:
My concern with using the default files is if you need to install on multiple computers you might have different versions of the files so I'd rather have the user download them to a specified directory, but I can be convinced this is not needed

mschwamb:
A quick note - conda installation hasn't been updated since assist is not conda installable - so the conflicts are because that effort has been stalled

mschwamb:
closing as I've taken the notes and put them in as tickets.

PR #662, opened 2023-09-28
Update README.md

mschwamb:
- [X] Does pip install still work?


astronomerritt (approved):
Seems good to me, approved. Out of interest, is this info going to be archived anywhere else? For example the section in the docs for contributors.

mschwamb:
I think if you're a contributor, you either will know this or you're going to be asking before putting in a pull request. I have the info copied. But we have lots of examples of what we need to do in the repo now

PR #667, opened 2023-09-28
Changed the code around initialization of t_picket.

### LINKED ISSUE #660: add docstrings and some comments to explain what's going on to ephemeris generation bits
I think it would be good to get the docstrings in to explain what's going on in the ephemeris generation functions. I don't think it will take very long but some quick description will be helpful to us in 6 months to a year.

### LINKED ISSUE #666: get rid of hard-coded number for initial t_picket value

matthewholman:
Fixes #666
Fixes #660 
This eliminates the hard-coded initial value for t_picket and includes some comments regarding why an initial value is needed.

### DIFF
--- a/src/sorcha/ephemeris/simulation_driver.py
+++ b/src/sorcha/ephemeris/simulation_driver.py
@@ -80,6 +80,7 @@ def create_ephemeris(orbits_df, pointings_df, args, configs):
     nside = 2 ** configs["ar_healpix_order"]
     first = 1  # Try to get away from this
 
+    ephemeris_csv_filename = None
     if args.output_ephemeris_file and args.outpath:
         ephemeris_csv_filename = os.path.join(args.outpath, args.output_ephemeris_file)
 
--- a/src/sorcha/ephemeris/simulation_driver.py
+++ b/src/sorcha/ephemeris/simulation_driver.py
@@ -21,7 +21,7 @@ out_csv_path = get_data_out_filepath("ephemeris_output.csv")
 
 
 def create_ephemeris(orbits_df, pointings_df, args, configs):
-    """ Generate a set of observations given a collection of orbits
+    """Generate a set of observations given a collection of orbits
     and set of pointings.
 
     This works by calculating and regularly updating the sky-plane
@@ -30,7 +30,7 @@ def create_ephemeris(orbits_df, pointings_df, args, configs):
     A dictionary with pixel indices as keys and lists of ObjIDs for
     those objects in each HEALPix tile as values.  One of these
     calculations is called a 'picket', as one element of a long picket
-    fence.  At present, 
+    fence.  At present,
 
     Given a specific pointing, the set of HEALPix tiles that are overlapped
     by the pointing (and a buffer region) is computed.  These the precise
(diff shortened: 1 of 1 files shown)


src/sorcha/ephemeris/simulation_driver.py
```diff
@@ -21,23 +21,71 @@
...
 
-    ephemeris_csv_filename = None
-
     if args.output_ephemeris_file and args.outpath:
         ephemeris_csv_filename = os.path.join(args.outpath, args.output_ephemeris_file)
 
-    t_picket = 2460000.5
-
-    ephem, gm_sun = create_assist_ephemeris(args)
+    ephem, gm_sun, gm_total = create_assist_ephemeris(args)
```
drewoldag, line 86: I think that this line, and the change on line 88 shouldn't be included in this PR. These are accidental inclusions from the other PR that is in progress now, correct?

drewoldag, line 86:
```suggestion
    ephem, gm_sun = create_assist_ephemeris(args)
```

src/sorcha/ephemeris/simulation_driver.py
```diff
@@ -21,23 +21,71 @@
...
     furnish_spiceypy(args)
-    sim_dict = generate_simulations(ephem, gm_sun, orbits_df)
+    sim_dict = generate_simulations(ephem, gm_sun, gm_total, orbits_df)
```
drewoldag, line 88:
```suggestion
    sim_dict = generate_simulations(ephem, gm_sun, orbits_df)
```

drewoldag (approved):
This looks fine to me, but in order to get the unit tests to pass, I think that you'll need to remove the references to `gm_total`. I've added a couple of "suggested changes" so you should be able to click "add suggestion to batch" on both of them, then commit them in a single commit to get the unit tests passing.

matthewholman (commented):
Thanks!

drewoldag:
Meg called out that 0 could potentially be a problem is some cases where observations in the 1970s are being considered. 

An alternative to using `t_picket = 0` would be to use the negative machine limit for integers like `t_picket = np.iinfo(np.int32)`. 

Using that approach the initial value for `t_picket = -2147483648`

matthewholman:
JD = 0 is January 1, 4713 BC.  I think we're ok.

src/sorcha/ephemeris/simulation_driver.py
```diff
@@ -21,20 +21,68 @@
...
+    """
+
     ang_fov = configs["ar_ang_fov"]
     buffer = configs["ar_fov_buffer"]
     picket_interval = configs["ar_picket"]
     obsCode = configs["ar_obs_code"]
     nside = 2 ** configs["ar_healpix_order"]
     first = 1  # Try to get away from this
 
-    ephemeris_csv_filename = None
```
drewoldag, line 31: I think removing this line is a problem. It will cause runtime errors if an output file isn't specified.

drewoldag (changes requested):
We need to add back in the `ephemeris_csv_filename = None` line.

drewoldag (approved):
I added the `ephemeris_csv_filename = None` line back in, so we should be good to go.

mschwamb:
Ah it's JD. I thought it was in MJD. 0 is fine. I'm going to merge in.

PR #675, opened 2023-09-29
Log and exit if complex parameter file wasn't provided

### LINKED ISSUE #578: Code does not fail gracefully when complex parameters input file expected but not given at command line
The light curve model and comet activity functionality requires parameters to be given in the complex parameters file, an optional argument on the command line.

If either `lc_model` or `comet_activity` is not None, the code will look for this complex parameters file - but if it doesn't exist, the code fails with a traceback:

```
  File "/opt/miniconda3/envs/sorcha/bin/sorcha", line 8, in <module>
    sys.exit(main())
  File "/Users/stephaniemerritt/Projects/sorcha/src/sorcha/sorcha.py", line 425, in main
    runLSSTPostProcessing(cmd_args, pplogger)
  File "/Users/stephaniemerritt/Projects/sorcha/src/sorcha/sorcha.py", line 136, in runLSSTPostProcessing
    reader.add_aux_data_reader(CSVDataReader(args.complex_parameters, configs["aux_format"]))
  File "/Users/stephaniemerritt/Projects/sorcha/src/sorcha/readers/CSVReader.py", line 39, in __init__
    self.header_row = self._find_header_line()
  File "/Users/stephaniemerritt/Projects/sorcha/src/sorcha/readers/CSVReader.py", line 68, in _find_header_line
    with open(self.filename) as fh:
FileNotFoundError: [Errno 2] No such file or directory: ''
```

This should instead provide a helpful message to the user and be logged in the error log.

However: I also wonder how likely it is that a user would create a light curve model that doesn't require any additional parameters... at the moment we have no way of accounting for that, as I've just submitted a PR request that makes the code look for the complex parameters file if `lc_model `is not None (as per #509). Thoughts welcome.

drewoldag:
Log and exit if user requested lightcurve or comet model, but no complex param file was provided.

Fixes #578 .

The logic is fairly straight forward here. But there was a question in the issue about whether it makes sense to enforce this or not - because it's possible that some user defined light curve or comet activity model may not require any additional parameters at all. 

I think it's a valid case to consider, but I can't immediately think of a simple way to enable that functionality.  I'll try to think a little more about this though. 

### DIFF
--- a/src/sorcha/sorcha.py
+++ b/src/sorcha/sorcha.py
@@ -480,6 +480,14 @@ def main():
         pplogger.error("ERROR: A+R simulation not enabled and no ephemerides file provided")
         sys.exit("ERROR: A+R simulation not enabled and no ephemerides file provided")
 
+    if configs["lc_model"] and cmd_args["complex_physical_parameters"] is None:
+        pplogger.error("ERROR: No complex physical parameter file provided for light curve model")
+        sys.exit("ERROR: No complex physical parameter file provided for light curve model")
+
+    if configs["comet_activity"] and cmd_args["complex_physical_parameters"] is None:
+        pplogger.error("ERROR: No complex physical parameter file provided for comet activity model")
+        sys.exit("ERROR: No complex physical parameter file provided for comet activity model")
+
     if "SORCHA_SEED" in os.environ:
         cmd_args["seed"] = int(os.environ["SORCHA_SEED"])
         pplogger.info(f"Random seed overridden via environmental variable, SORCHA_SEED={cmd_args['seed']}")


astronomerritt:
I chatted with @bernardinelli about this on the Slack a while back and the conclusion was that a user would have to be doing something quite silly and awkward to create a lightcurve or activity model that didn't require any additional parameters, so it's absolutely fine to go ahead and assume they will always be supplying the complex parameters file.

We can say in the docs that the complex parameters file is absolutely required if using a lightcurve or activity model, and if the user really wants to be silly they can supply a dummy file.

mschwamb (approved):

PR #683, opened 2023-09-29
Issue #607: Removing separatelycsv output option.

### LINKED ISSUE #607: Remove separatelycsv option for output_format as an option
Now that we have ephemeris generation that produces huge files we really do not want to let any user create output files for each object. It's just dangerous given the ephemeris file is 1.9 Gb in total for 1E4 objects, separate files is  not a good idea.  So we should get rid of the option so the user can't do something silly here. If the user wants separate files they can split up their objects into separate input files and it will do the same thing. 

https://github.com/dirac-institute/sorcha/blob/main/src/sorcha/modules/PPOutput.py#L146C-L159C22

astronomerritt:
Fixes #607.

Removed 'separatelycsv' as an output option from PPOutput.py and also removed the related lines from the unit tests.

### DIFF
--- a/demo/OIFconfig_benchmark.ini
+++ b/demo/OIFconfig_benchmark.ini
@@ -135,8 +135,8 @@ SSP_maximum_time = 0.0625
 # -u or --outfile               Output file path.
 # -t or --stem                  Output file stem.
 
-# Output format. The 'separatelycsv' option saves output into separate CSVs for each object.
-# Options: csv, separatelycsv, sqlite3, hdf5
+# Output format.
+# Options: csv, sqlite3, hdf5
 output_format = csv
 
 # Size of output. Controls which columns are in the output files.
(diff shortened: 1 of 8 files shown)


mschwamb (changes requested):
What about  [PPConfigParser.py](https://github.com/dirac-institute/sorcha/blob/3427d6b4f205144cbe015c655f4299a2273d669a/src/sorcha/modules/PPConfigParser.py#L628)


   `if config_dict["output_format"] not in ["csv", "separatelycsv", "sqlite3", "hdf5", "h5"]: in `

astronomerritt:
Oops, good catch. One sec.

mschwamb (changes requested):
Looks good. There are a few mentions of separatelycsv in some config files can you strip those as well?

astronomerritt:
Agh. I knew I should have grepped. Apologies. That should be all of them now.

Not sure why codecov is complaining again.

astronomerritt:
Codecov is failing due to an untested function in PPConfigParser.py - log_error_and_exit. This is a very basic function and I'm honestly not sure how I'd unit test it. Probably okay to go ahead and approve/merge.

mschwamb (approved):
Looks good

PR #698, opened 2023-10-06
remove python 3.8

### LINKED ISSUE #697: smoke test failing
pytables not making wheels for python3.8 anymore.

mschwamb:
move to python 3.11 instead of 3.8 since for tests and installation because of pytables no longer making wheels for python 3.8

Fixes #697


drewoldag (approved):
This looks correct to me.

PR #714, opened 2023-10-25
Changing code to use midpoint of observation.

### LINKED ISSUE #590: convert the fieldMJD to be the mid-point of the visit for rubin_sim inputs
We are using observationStartMJD so we should adjust this half the visit time I think looking at the [schema](https://rubin-sim.lsst.io/rs_scheduler/output_schema.html) . It may change in actual operations, but it's how rubin_sim works See discussion in #586

astronomerritt:
Fixes #590.

When the code is run with 'lsst' as the survey name (which is default), `PPReadPointingDatabase()` will create a new column named `observationMidpointMJD_TAI`, which is `observationStartMJD_TAI` + `visitTime`/2/86400 (to convert it into days). The ephemeris generation has been changed to use this new column for all calculations.

Note that I needed to update the demo pointing directory `./demo/baseline_v2.0_1yr.db` because it didn't include the `visitTime` column.

I have also updated all of the tests, and added a test to make sure the midpoint is being correctly calculated. 

Note that many of the unit tests use an old test version of the pointing database, `./tests/data/baseline_10klines_2.0.db`, that also doesn't include the `visitTime` column. In order to save a lot of tedious rewriting, `PPReadPointingDatabase()` will revert to its old behaviour if the survey name is "test". At some point, these tests should be rewritten to use `./demo/baseline_v2.0_1yr.db`.


mschwamb:
`Note that many of the unit tests use an old test version of the pointing database, ./tests/data/baseline_10klines_2.0.db, that also doesn't include the visitTime column. In order to save a lot of tedious rewriting, PPReadPointingDatabase() will revert to its old behaviour if the survey name is "test". At some point, these tests should be rewritten to use ./demo/baseline_v2.0_1yr.db.`

I think this is dangerous to do. We could update baseline_v2.0_1yr.db.in the repo and also check whether or not the older databases have this information and how to adjust to those older databases

astronomerritt:
Honestly I agree, but I'm not sure what the best thing to do is. 

I've updated baseline_v2.0_1yr.db.in the repo already - it didn't have visitTime because when I made it, I only made it with the "necessary" columns to keep the file size down. I expect the same thing is true for baseline_10klines_2.0.db. 

So it isn't that the older databases don't have visitTime, it's just that these specific smaller test databases were created without it.

I could always just edit a visitTime column into baseline_10klines_2.0.db.

astronomerritt:
So it turns out that not only was `/data/baseline_10klines_2.0.db` indeed just the first 10k lines of the baseline_v2.0 database... I _was_ the one who made it. Months ago. Because we didn't know where the original test pointing database came from.

Anyway, I have added the `visitTime` column to `/data/baseline_10klines_2.0.db`, removed that `surveyname == "test"` condition in `PPReadPointingDatabase()`, and updated all tests accordingly. This should be good to go now.

mschwamb (approved):

PR #728, opened 2023-11-10
doc updates

mschwamb:
make changes to make it clearer for development install instructions


astjoephysics (commented):
Pip install works as expected with this command

astjoephysics (approved):

PR #767, opened 2024-01-08
Proper update of logging incl. test files

### LINKED ISSUE #682: Change all imports of pplogger to use the pplogger stored in args
So pplogger, the object which controls the log files, is now stored in the sorchaArguments object args, which is very cool! But in most of the code, the pplogger object is still being reinitialised via:

```
import logging
...
pplogger = logging.getLogger(__name__)
```

This works, but it's pointless (given that we can now access it via args) and inconsistent.

It would be nice to go through the code and remove all of this in favour of passing args (if it's not already being passed) and calling args.pplogger.info or args.pplogger.error.

This is obviously not high priority.

astjoephysics:
Fixes #682 .

Describe your changes.
Made logging consistent with Python logger by taking logging out of sorchaArguments. Updated tests accordingly.


mschwamb (approved):
Looks good to be. Made a small change to put the verbosity of the log file back to what was in the original unit test

PR #778, opened 2024-01-11
Adding notebooks to generate Sorcha runscripts.

astronomerritt:
Added two notebooks to `./docs/notebooks/`: demo_GenerateBashScripts and demo_GenerateSLURMScripts. These can be modified and used to generate shell or SLURM runscripts for Sorcha. Updated the README.md accordingly.


mschwamb:
@Gerenjie  does this in a different way, where there is one slurm command/file that gets iterated over. 

On the command line: sbatch --array=0-130 sorcha_mpcorb.sh

mpcorb.sh has in it 
#!/bin/bash
#SBATCH --job-name=sorcha_mpcorb   # Job name
#SBATCH --mail-type=FAIL            # Mail events (NONE, BEGIN, END, FAIL, ALL)
#SBATCH --mail-user=youremail@someplace.edu   # Where to send mail
#SBATCH --ntasks=1                  # Run a single task
#SBATCH --mem=16gb                   # Job Memory
#SBATCH --time=03:00:00             # Time limit hrs:min:sec
#SBATCH --output=logs/mpcorb-%a.log    # Standard output and error log

sleep ${SLURM_ARRAY_TASK_ID}
sleep ${SLURM_ARRAY_TASK_ID}
date
sorcha -c ./survey_setups/Rubin_full_footprint.ini -p ./dp031/mpcorb/infiles/mpcorb_colors_${SLURM_ARRAY_TASK_ID}.csv -ob ./dp031/mpcorb/infiles/mpcorb_orbits_${SLURM_ARRAY_TASK_ID}.csv -pd ./baseline_v3.3_10yrs.db -o ./dp031/mpcorb/outfiles/ -t mpcorb_${SLURM_ARRAY_TASK_ID} -f
date

Is this a simpler option that's in the notebook?

astronomerritt:
I could probably implement something simpler along those lines (although Jake's version uses the command line and sbatch, which couldn't be done in a notebook) - as it is, I just made changes to the notebook I was already using to generate Kelvin scripts.

mschwamb:
We should people the best/simplest example. Can you test if what Jake does can work on Kelvin?

astronomerritt:
It would definitely work, it's a standard way of using sbatch (though it requires a very strict naming scheme, so to test it I'd need to mess about a bit), but it would create each run of Sorcha as a single job. If you're running ~1500 instances of Sorcha at once (i.e. for DP0.3) that's going to lead to a lot of created jobs (and a lot of emails).

My script packages a number of Sorcha runs into a smaller number of jobscripts. So if one wants to run Sorcha 1500 times, it can separate that up into batches of 100 and outputs 15 jobscripts, each of which runs Sorcha 100 times over 15 cores. To me this is a neater way of running things.

Jake's solution is good and should probably be a page in the docs. Up to you whether my notebook is still useful!

mschwamb (approved):

PR #780, opened 2024-01-15
Adding command to copy config files.

### LINKED ISSUE #776: Package config files
Package config files so that they can be copied from install into current directory as a default config file for the user

astronomerritt:
Fixes #776 .

There is now a command (`sorcha_copy_configs`) which will copy the example configuration files from where-ever Sorcha is installed on the user's machine to a location of their choosing (or, by default, the current working directory). This will work for pip and Conda installations.

The user can also choose (via command-line flag) whether to copy the Rubin config files, the demo config file, or all of them. Usage:

`sorcha_copy_configs [-h] [-f FILEPATH] [-c [{rubin,demo,all}]]`

Updated pyproject.toml, also wrote a unit test.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,6 +41,7 @@ sorcha = "sorcha.sorcha:main"
 makeSLURMscript = "sorcha.utilities.makeSLURMscript:main"
 createResultsSQLDatabase = "sorcha.utilities.createResultsSQLDatabase:main"
 bootstrap_sorcha_data_files = "sorcha.utilities.retrieve_ephemeris_data_files:main"
+sorcha_copy_configs = "sorcha.utilities.sorcha_copy_configs:main"
 
 [project.urls]
 "Documentation" = "https://sorcha.readthedocs.io/en/latest/"
--- a/src/sorcha/utilities/sorcha_copy_configs.py
+++ b/src/sorcha/utilities/sorcha_copy_configs.py
@@ -8,7 +8,7 @@ from sorcha.modules.PPConfigParser import PPFindDirectoryOrExit
 
 
 def copy_demo_configs(copy_location, which_configs):
-    """ "
+    """
     Copies the example Sorcha configuration files to a user-specified location.
 
     Parameters
@@ -66,7 +66,7 @@ def main():
     usage: sorcha_copy_configs [-h] [-f FILEPATH] [-c [{rubin,demo,all}]]
         arguments:
           -h, --help                                             Show this help message and exit.
-          [-f FILEPATH, --filename FILEPATH]                       Filepath where you want to copy the config files. Default is current working directory.
+          [-f FILEPATH, --filename FILEPATH]                     Filepath where you want to copy the config files. Default is current working directory.
           [-c [{rubin,demo,all}]],--configs [{rubin,demo,all}]   Which config files you want: options are "rubin", "demo" and "all". Default is "all".
     """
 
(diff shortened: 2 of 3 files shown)


mschwamb:
From the sorcha calls it was discussed that it would be good to make this interactive so the user can see the options and select which config file they want or if they want them all and then confirm their answer through the keyboard

astronomerritt:
Changes:

- The user now interacts with the command line to select which config files they want, instead of using command line flags, by following prompts printed to the terminal.
- Added a function that parses and error handles the user input.
- Wrote a unit test to cover the above function.

Let me know if I should add more explanatory text to the information printed to the terminal.

mschwamb:
Likely the user only wants 1 rubin config file. Let's give them the option to download each one

mschwamb:
where the config files get copied to can be a config variable

astronomerritt:
Changes made:

- The user now specifies the location to which the files are copied on the command line using the -f flag. The default is the current directory. This brings it in line with the cache script.
- The utility now allows the user to select either the Rubin circular approximation or the Rubin full footprint config file.

mschwamb (changes requested):
raceback (most recent call last):
  File "/Users/mschwamb/mambaforge/envs/sorcha/bin/sorcha_copy_configs", line 8, in <module>
    sys.exit(main())
  File "/Users/mschwamb/Research/python/lsst/survey_simulator/sorcha/sorcha_main_repo/sorcha/src/sorcha/utilities/sorcha_copy_configs.py", line 112, in main
    parser.add_argument(
NameError: name 'parser' is not defined

astronomerritt:
Oops. Fixed.

mschwamb:
Let's take the demo config file. We are packaging the demo files. The user can grab the demo config file from that

mschwamb:
Which configuration file(s) would you like to copy?:

1. Rubin-specific configuration file using circular approximation of camera footprint (faster).

2. Rubin-specific configuration file using full camera footprint (slower, but more accurate).

3. All.

Please enter a number and hit Return/Enter.
4
  File "/Users/mschwamb/mambaforge/envs/sorcha/bin/sorcha_copy_configs", line 8, in <module>
    sys.exit(main())
  File "/Users/mschwamb/Research/python/lsst/survey_simulator/sorcha/sorcha_main_repo/sorcha/src/sorcha/utilities/sorcha_copy_configs.py", line 129, in main
    which_configs = parse_file_selection(file_select)
  File "/Users/mschwamb/Research/python/lsst/survey_simulator/sorcha/sorcha_main_repo/sorcha/src/sorcha/utilities/sorcha_copy_configs.py", line 83, in parse_file_selection
    which_configs = selection_dict[file_select]
KeyError: 4

mschwamb:
It will copy over a file if I already have it in the directory. Would be good to add a -f flag and force people to double check that they want us to overwrite on a file they already have.

mschwamb:
Maybe not a -f but something that says yeah overwrite what I currently have in the directory

mschwamb:
I'd follow what the download script is doing
  --cache CACHE  Local directory where downloaded files will be stored.
  -f, --force    Delete and re-download data files.

astronomerritt:
Makes sense. I may go for `--path PATH` and then `-f --force`, if that suits.

mschwamb:
Whatever works

mschwamb (approved):

Issue #758, opened 2023-12-29
Should this line in PPAddUncertainties by a np.sqrt?


mschwamb:
sqrt is undefined I think. I believe this line never gets executed in sorcha, but I thought worth a check: 

https://github.com/dirac-institute/sorcha/blob/084172ab265ab3f84f5b546fd6c3b14ce7933f64/src/sorcha/modules/PPAddUncertainties.py#L219

mschwamb:
Closed by PR #813

Issue #600, opened 2023-09-08
ccdVisitID


Gerenjie:
Although the CCD on which chip each detection falls is calculated at some point, it is not outputted. Having this as an (optional) output would allow us to properly populate the ccdVisitID column the DiaSource table.

mschwamb:
@samncorn  has this functionality already in the camera CCD association function but we don't have this enabled. First we need to check that the naming scheme is the same for the real LSSTCam.

samncorn:
I think the current implementation is incorrect, so until I can verify it the values returned for which detector a detection is on should be disregarded.

PR #813, opened 2024-02-09
Issue 752

### LINKED ISSUE #752: Error logging/die gracefully if the default config file does not exist or reading in the user provided camera detector file fails
Now that we have a don't have error handling on if reading in our default built-in installed config file fails or reading in the user provided one fails. We check the pathway exists for the user provided detector file, but no in the case we use the built-in version. 

https://github.com/dirac-institute/sorcha/blob/f7ae7ae7f429c9afe28c02557c4d31c1499e427a/src/sorcha/modules/PPFootprintFilter.py#L397C1-L405C47

samncorn:
Fixes #752, #758 .

Added check for if default camera file exists.

Fixed a call to a sqrt function that didn't exist.

### DIFF
--- a/src/sorcha/modules/PPAddUncertainties.py
+++ b/src/sorcha/modules/PPAddUncertainties.py
@@ -272,7 +272,7 @@ def calcAstrometricUncertainty(
     error_rand = calcRandomAstrometricErrorPerCoord(FWHMeff, SNR, astErrCoeff)
     # random astrometric error for nvisit observations
     if nvisit > 1:
-        error_rand = error_rand / sqrt(nvisit)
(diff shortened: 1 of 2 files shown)


mschwamb:
Why does segmented_area() exist. It's not something in the repo currently. I don't see it being used in PPFootprintFilter.py. Is there a reason to include it?

mschwamb:
Same thing for radec_to_focal_plane(). It does not look like it's being used.

samncorn:
My bad, I was trying to clean stuff up to make #600 easier, forgot to make a separate branch for it. I'll remove those.

mschwamb (approved):

PR #822, opened 2024-02-15
Printing/testing demo command.

### LINKED ISSUE #692: add a sorcha demo command
add a sorcha demo command which would spit out the commands to run the demo.

### LINKED ISSUE #815: Have a everything works as supposed utility script check
I also think having a command (python script) that can compare the output if the seed is set and unsets the seed and compares the output to check everything is hunky-dory would be good. Looks like https://pymotw.com/2/filecmp/ could compare the files.

astronomerritt:
Fixes #692 .

- There is now a demo command that spits out the command which runs Sorcha from the demo files. It can be run by running `sorcha_demo_command`.
- This command also tells the user how to run `sorcha_copy_demo_files` to get the demo files.
- I'm not sure there's much point in a unit test for this function, as it only outputs text to the terminal, but I wrote one anyway so CodeCov stays happy.
- `sorcha_copy_demo_files` now also prints the demo command to the terminal.
- There is a function in sorcha_demo_command.py called `get_demo_command()` which will simply return the demo command as a string. If the demo command ever changes, this is the only place in the code that needs to be updated.

Fixes #815 (I think).

- There is now a test that obtains the current version's working demo command and runs it on the command line to ensure that it both produces output and has an empty error log.

### DIFF
--- a/tests/sorcha/test_sorcha_demo_command.py
+++ b/tests/sorcha/test_sorcha_demo_command.py
@@ -1,13 +1,16 @@
 def test_copy_demo_command(capsys):
-    from sorcha.utilities.sorcha_demo_command import print_demo_command
+    from sorcha.utilities.sorcha_demo_command import print_demo_command, get_demo_command
 
     print_demo_command(printall=True)
+    current_demo_command = get_demo_command()
 
     captured = capsys.readouterr()
 
     expected = (
         "\nThe command to run the Sorcha demo in this version of Sorcha is:\n\n"
-        + "\033[1;32;40msorcha -c sorcha_config_demo.ini -p sspp_testset_colours.txt -ob sspp_testset_orbits.des -pd baseline_v2.0_1yr.db -o ./ -t testrun_e2e\033[0m\n\n"
+        + "\033[1;32;40m"
+        + current_demo_command
+        + "\033[0m\n\n"
         + "WARNING: This command assumes that the demo files are in your working directory.\n\n"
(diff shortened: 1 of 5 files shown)


tests/sorcha/test_sorcha_demo_command.py
mschwamb, line 1: This only tests that the output from the function is correct. If the command line arguments were changed  this would not catch it. 
Can you use os.system() and check the error file is empty and there is output?

mschwamb (changes requested):
I think there should be a function that returns back a string of just the demo command line to run. Then I think it would be possible to use os.system to see if it successfully runs.

tests/sorcha/test_sorcha_demo_command.py
astronomerritt: Yep, I know - I haven't done the second part yet, I was getting this out of the way first.

astronomerritt:
Agreed. I'll get that done tomorrow and add it to this PR then.

astronomerritt:
I have no idea why CodeCov is showing such a dramatic dip in project coverage - if anything it should be increasing now that we're covering `main()` in sorcha.py.

astronomerritt:
Ah. CodeCov has stopped ignoring citation_text.py. I believe this is because I'm now running sorcha.py as part of a unit test, which imports `cite_sorcha` - CodeCov can ignore a file, but it's not ignoring the function if it's imported by something else. Not sure how to fix this just yet.

mschwamb (approved):

PR #843, opened 2024-02-29
Issue 821

### LINKED ISSUE #821: RARate(deg/day) column is actually RA rate * cos(Dec)
The column labelled RARate(deg/day) is actually RA rate * cos(dec) as calculated by `calculate_rates_and_geometry` in ephemeris generation.

However, `PPTrailingLoss` assumes that this column is simply the RA rate and then calculates RA rate * cos(Dec) again.

samncorn:
Fixes #821 

Trailing losses now assume that the angular motion provided is not scaled to declination. That is, the cos(dec) term has either been applied, or the value was obtained in some manner that obtains the same effect.


mschwamb:
I will approve this one the fastmover stuff is merged in. Mainly trying to not implement changes that will impact determining if the fastmover changes are okay

PR #857, opened 2024-03-06
use different config file for cprofiler

### LINKED ISSUE #856: benchmark script does not work
It's using the wrong config file. It's not using the wrong config file

mschwamb:
If I don't make this change, I get an error about the format of the input files and the profiler does not actually run.

Fixes #856


astronomerritt (approved):
Weird. I could have sworn I changed that - it ran on my machine, after all. Thanks for spotting it!

PR #863, opened 2024-03-13
providing exposure time to trailing loss calculations

### LINKED ISSUE #849: Trailing losses assumes 30s exposures but LSST observations could be longer or shorter
Low solar elongation near-sun twilight survey observations are not 30s but shorter I believe. 

The trailing loss calculation assumes that the exposures are 30s, but not all LSST observations are 30s. 

calcTrailingLoss takes a default value but  PPTrailingLoss does not send the exposure time of the observations.

samncorn:
Fixes #849.

Modified TrailingLoss and AddUncertainty to provide the 'visitExposureTime' columns to calcTrailingLoss.

Added 'visitExposureTime' columns to the relevant unit tests.


src/sorcha/modules/PPTrailingLoss.py
mschwamb, line 1: https://github.com/dirac-institute/sorcha/blob/3279da4cd19954917f445b0b3639d53ac3f431c1/src/sorcha/modules/PPTrailingLoss.py#L50C1-L51C55 - doc string needs to be changed to say float or array of floats

mschwamb (changes requested):
minor change in doc string. otherwise looks

src/sorcha/modules/PPAddUncertainties.py
```diff
@@ -191,7 +196,10 @@ def uncertainties(
 
     if configs.get("trailing_losses_on", False):
         dMag = PPTrailingLoss.calcTrailingLoss(
-            detDF[dra_name] * degCos(detDF[dec_name]), detDF[ddec_name], detDF[seeingName]
```
mschwamb, line 194: On other change:

AstRARate(deg/day) is RARateCosDec_deg_day in the new ephemeris generation so you shouldn't be multiplying be cosdec here 

https://github.com/dirac-institute/sorcha/blob/3279da4cd19954917f445b0b3639d53ac3f431c1/src/sorcha/modules/PPTrailingLoss.py#L88C1-L89C48

mschwamb (approved):

PR #879, opened 2024-03-22
Update .readthedocs.yml

mschwamb:
add ability for PDF to be created from read the docs


astronomerritt (approved):
That's a lot easier than I thought it would be.

astjoephysics (approved):

PR #887, opened 2024-04-08
Add more checks and debugging strings into file readers

### LINKED ISSUE #854: Bad error message for incorrect aux file delimiter
When you run Sorcha with the config file specifying the wrong auxiliary file, the first thing to break is the observing filter matching, so you get the following error message which is unclear and unhelpful (I have gotten this dozens of times while running Sorcha myself):
ERROR: PPGetMainFilterAndColourOffsets: colour offset columns in physical parameters file do not match with observing filters specified in config file.

It would be nice to get a message saying "aux files don't appear to be separated by [comma/space]" instead.

jeremykubica:
Fixes #854 

Adds more (and earlier) checks that a file is being read in with the correct format and delimiter. Produces more meaningful error messages when a mismatch is seen, including outputting the header line to the info logs.

The automatic formatting run picked up a few additional small changes that I included.


mschwamb (approved):
Looks great to me. Thanks @jeremykubica

PR #923, opened 2024-04-23
Adding more switches to config file.

### LINKED ISSUE #897: Add config file variable to turn off randomizing the astrometry and photometry
Add config file variable to turn off randomizing the astrometry and photometry

### LINKED ISSUE #899: Add a hidden camera NONE option
allows you to turn off the camera footprint filter for debugging purposes or for other reasons if you know what you're doing

### LINKED ISSUE #901: hidden config variable to turn vignetting off
hidden config variable to turn vignetting off

astronomerritt:
Fixes #901.
- There is now a hidden config key in the [EXPERT] section, "vignetting_on", which can be used to turn vignetting on and off. By default if the key is not included, this will be set to True.

Fixes #899.
- Setting "camera_model" to "none" in the config file results in no camera model being applied.

Fixes #897.
- There is now a hidden config variable in the [EXPERT] section, "randomization_on", which can be used to turn randomization on and off. By default if the key is not included, this will be set to True.

For all of the above:
- Relevant messages are printed to the log to let the user know what is happening.
- Unit tests have been updated, including testing the expected behaviour when randomisation and camera model are turned off.
- A notebook has also been updated.

### DIFF
--- a/src/sorcha/sorcha.py
+++ b/src/sorcha/sorcha.py
@@ -261,8 +261,8 @@ def runLSSTSimulation(args, configs):
             verboselog(
                 "NOTE: new columns RATrue_deg and DecTrue_deg are EQUAL to columns RA_deg and Dec_deg."
             )
-            observations["RATrue_deg"] = observations["RA_deg"]
-            observations["DecTrue_deg"] = observations["Dec_deg"]
+            observations["RATrue_deg"] = observations["RA_deg"].copy()
+            observations["DecTrue_deg"] = observations["Dec_deg"].copy()
 
         verboselog("Applying field-of-view filters...")
         verboselog("Number of rows BEFORE applying FOV filters: " + str(len(observations.index)))
--- a/src/sorcha/sorcha.py
+++ b/src/sorcha/sorcha.py
@@ -222,15 +222,22 @@ def runLSSTSimulation(args, configs):
         else:
             observations["PSFMagTrue"] = observations["trailedSourceMagTrue"]
 
-        verboselog("Calculating effects of vignetting on limiting magnitude...")
-        observations["fiveSigmaDepth_mag"] = PPVignetting.vignettingEffects(observations)
+        if configs["vignetting_on"]:
+            verboselog("Calculating effects of vignetting on limiting magnitude...")
+            observations["fiveSigmaDepth_mag"] = PPVignetting.vignettingEffects(observations)
+        else:
+            verboselog(
+                "Vignetting turned OFF in config file. 5-sigma depth of field will be used for subsequent calculations."
+            )
+            observations["fiveSigmaDepth_mag"] = observations["fieldFiveSigmaDepth_mag"]
 
         # Note that the below code creates trailedSourceMag and PSFMag
         # as columns in the observations dataframe.
         # These are the columns that should be used moving forward for filters etc.
         # Do NOT use trailedSourceMagTrue or PSFMagTrue, these are the unrandomised magnitudes.
         verboselog("Calculating astrometric and photometric uncertainties...")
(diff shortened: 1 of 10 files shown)


tests/sorcha/test_PPApplyFOVFilter.py
```diff
@@ -92,7 +92,7 @@ def test_PPApplyFOVFilters():
     configs = {
         "camera_model": "footprint",
         "footprint_path": get_test_filepath("detectors_corners.csv"),
-        "footprint_edge_threshold": 0.0,
+        "footprint_edge_threshold": 10.0,
```
mschwamb, line 95: Why is the footprint edge on here? There is a test that checks this I thought separately?

src/sorcha/sorcha.py
```diff
@@ -244,10 +251,13 @@ def runLSSTSimulation(args, configs):
...
         )
 
-        verboselog("Randomising astrometry...")
-        observations = PPRandomizeMeasurements.randomizeAstrometry(
-            observations, args._rngs, sigName="astrometricSigma_deg", sigUnits="deg"
-        )
+        if configs["randomization_on"]:
+            verboselog("Randomizing astrometry...")
+            observations = PPRandomizeMeasurements.randomizeAstrometry(
+                observations, args._rngs, sigName="astrometricSigma_deg", sigUnits="deg"
```
mschwamb, line 257: We want to make sure the columns added by randomize Astrometry are added I think but just the values are the same as the one calculated

mschwamb (changes requested):

tests/sorcha/test_PPApplyFOVFilter.py
astronomerritt: To make sure that this test actually does something. I noticed while I was running this test that the tested objects _all_ make it through the footprint filter. Setting the edge threshold to high at least removes a couple of them and ensures that the footprint filter is activating properly with the right config keys, which is all that is being tested here.

mschwamb: But testing the foorprint edge is a different test, right? Shouldn't this be testing the generic part of the footprint fitler?

src/sorcha/sorcha.py
astronomerritt: Would something like the below work in sorcha.py or would you prefer I did this inside the function:

```
        if configs["randomization_on"]:
            verboselog("Randomizing astrometry...")
            observations = PPRandomizeMeasurements.randomizeAstrometry(
                observations, args._rngs, sigName="astrometricSigma_deg", sigUnits="deg"
            )
        else:
            verboselog("Randomization turned off in config file. No astrometric randomization performed.")
            observations["RATrue_deg"] = observations["RA_deg"]
            observations["DecTrue_deg"] = observations["Dec_deg"]

```

mschwamb: something like this but you might need to explicitly copy.

tests/sorcha/test_PPApplyFOVFilter.py
astronomerritt: I have brought the tests in PPApplyFOVFilter in line with the tests on its child functions: so now they all test exactly the same data and give exactly the same results. The part that tests "footprint" was copied directly from Sam's test_PPApplyFootprint.py.

Turns out I wrote the tests for PPSimpleSensorArea and PPCircleFootprint, so I double-checked them while I was there. All good and as expected. I've just standardised them so all tests use the first ten rows of the same data file.

src/sorcha/sorcha.py
astronomerritt: Done.

mschwamb (approved):

PR #945, opened 2024-05-11
sorcha: command line interface rework

mjuric:
Rework command line interface to follow (git-like) `sorcha <verb>` semantics (and speed it up significantly).  Presently available commands include:

    sorcha init      # downloads the ephemerides files
    sorcha run       # runs the simulation
    sorcha demo      # commands related to running the demo
    sorcha config    # copying/manipulating configuration files
    sorcha outputs   # post-processing simulation outputs

Running sorcha --help will show all available verbs (and each verb can be further queried with --help as well).

Implementation:

* The command-line entrypoints have all been moved to a separate package, named `sorcha_cmdline`.  This allows us to run the command line utility without initially importinhg the main sorcha package.  This, in turn, lets us exit quickly and print the help/error message (in case there was a mistake on the command line).  Importing sorcha can take 5 seconds or more, and making the user wait that long just to print out an error message is poor user experience.

* Old entry-points were all renamed to `sorcha-foo-bar`. For example, we now have `sorcha-init` (created from `sorcha_cmdline.init`'s `main` function). Specifically, this is what pyproject.toml now holds:

    [project.scripts]
    sorcha = "sorcha_cmdline.main:main"
    sorcha-run = "sorcha_cmdline.run:main"
    sorcha-config = "sorcha_cmdline.config:main"
    sorcha-demo = "sorcha_cmdline.demo:main"
    sorcha-outputs = "sorcha_cmdline.outputs:main"
    sorcha-init = "sorcha_cmdline.init:main"

  To add a new subcommand, follow this same pattern (add the python file with argument parsing to sorcha_cmdline module, and add an entry to [project.scripts].  be careful not to import `sorcha` until after the arguments have been parsed.).

* A generic driver named 'sorcha' searches for executables of the form `sorcha-<foo>` on the path.  These will be invoked when `sorcha <foo>` is invoked. While the user can also call these directly with sorcha-foo, invoking them using the `sorcha <verb>` syntax is more future-proof (e.g., it will continue working if we ever decide to implement a verb internal to the sorcha command, and not as a separate sub-command).

* Note that this allows third-party extension packages to add command-line utilities into sorcha's namespace (e.g., if an addon package named `des` installed a utility `sorcha-des` into the path, it would become automatically available as `sorcha des`). This follows the pattern popularized by git.

Fixes # .

Describe your changes.


mschwamb:
I think there needs to be much more discussion before this PR can be merged.  I thought the changes were going to be we're switching to sorcha_copy_demo_files to sorcha-copy-demo-files not refactoring to what git does. I have some concerns on the names and the functionality here.  calling something sorcha init sounds like one has to do that everytime before running a simulation which is not true.  Discusing this on a sorcha call would be useful

mschwamb:
change init to bootstrap

src/sorcha_cmdline/main.py
```diff
@@ -0,0 +1,67 @@
...
+    # Discover available sorcha verbs
+    available_verbs = find_sorcha_verbs()
+
+    if not available_verbs:
+        print("Error: No available 'sorcha-' utilities found.")
+        sys.exit(1)
+
+    # Set up the argument parser with epilog text
+    description = 'Sorcha survey simulator suite.'
+    epilog_text = "These the most common sorcha verbs:\n\n" \
```
astronomerritt, line 31: Typo: these are

src/sorcha_cmdline/main.py
```diff
@@ -0,0 +1,67 @@
...
+                  "   run       Run a simulation\n" \
+                  "   demo      Run a demo\n" \
```
astronomerritt, line 33: I understand that these descriptions are written to be general, in case we want to add more commands later. However, I think "run a demo" here is misleading. It makes it sound like the command runs the demo in the same way `sorcha run` would run the simulation. Maybe "Set up a demo simulation"?

src/sorcha_cmdline/main.py
```diff
@@ -0,0 +1,67 @@
...
+                  "   config    Prepare or manipulate configuration files\n" \
+                  "   outputs   Manipulate/package sorcha outputs\n" \
```
astronomerritt, line 35: Is init/bootstrap omitted from this list on purpose? Fine if it is, just checking.

src/sorcha_cmdline/run.py
```diff
@@ -0,0 +1,197 @@
+#
+# The `sorcha run` subcommand implementation
+#
+import argparse
+
+def main():
+    parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
```
astronomerritt, line 7: Is there any help text for this subcommand anywhere? I know it's fairly self-explanatory, but I wonder do we want a couple of words like "run main Sorcha simulation".

astronomerritt:
I've left some comments on the PR re: usability (and a typo).

In general though I am strongly in favour of this change!

mschwamb:
@mjuric We're happy for you to implement this - Can you address @astronomerritt's feedback?

src/sorcha_cmdline/main.py
mjuric: Agreed, fixed!

src/sorcha_cmdline/main.py
mjuric: Fixed.

src/sorcha_cmdline/main.py
mjuric: Added "bootstrap Download datafiles required to run sorcha"

src/sorcha_cmdline/config.py
```diff
@@ -0,0 +1,88 @@
...
+# sorcha config
+#
+
+def main():
+    # Create the top-level parser
+    parser = argparse.ArgumentParser(prog="sorcha-config", description="Sorcha Configuration Utility")
+    subparsers = parser.add_subparsers(title="commands", description="Available commands", help="Command to execute", dest='command')
+
+    # Add the `prepare` subcommand
+    prepare_parser = subparsers.add_parser("prepare", help="Prepare configuration")
```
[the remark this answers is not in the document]
mjuric: I'm in favor of prepare because I'm thinking of it as preparing a configuration file to run some kind of a simulation. Whether it's generated or fetched from somewhere is irrelevant to the user -- what it does from their perspective is preparing the environment (config, and possibly other things in the future -- like data files) for a sorcha run.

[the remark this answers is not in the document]
mschwamb: can we add a "prepare default configuration"  or "prepare default configuration"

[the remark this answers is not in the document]
mjuric: Hmm, there's no true "default" -- there are already two configs to choose from (and as more instruments are added, more may appear).

Thinking more about this, how about using `init` here? I.e.: `sorcha config init` initializes the configuration files for a simulation run.

Even better, I can drop `config` (i.e., only have `sorcha init`). This would make a typical workflow look like:
```
sorcha init
... fiddle with config files ...
sorcha run
```
I like this latter option (also saves 5 characters of typing :) )...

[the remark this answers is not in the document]
mschwamb: I like init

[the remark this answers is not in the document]
mjuric: 👍. Give me five minutes...

mjuric:
@astronomerritt @mschwamb: assuming the unit tests pass, I think this is ready to be merged; take a look.

mschwamb:
Yeah.- I think this looks good. This PR rocked my world... literally looking at it when the earthquake waves passed by...

mschwamb:
@mjuric  While this fresh in your head, can you write a paragraph on why you did this - what it improves so we can add it to the sorcha paper? I'm procrastinating from parsing my planet 9 sorcha sims output

mschwamb (approved):

astronomerritt (approved):

astronomerritt:
Looks great to me. Thanks for making the changes!

Issue #978, opened 2024-07-12
Command line argument naming should follow UNIX conventions


mjuric:
Unix programs these days typically follow the [GNU conventions](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html) for command line arguments. We deviate from them in two ways:

* Have short arguments with more than one character (e.g. `-cp` in `sorcha run`). This is ambiguous with shortening `-c -p` as `-cp` (see the second bullet [here](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html)), and makes `argparse` disable this feature entirely.
* Have long arguments with underscores (`_`). This goes against GNU conventions (quote: "Long options consist of -- followed by a name made of alphanumeric characters and dashes."), but more importantly it makes our code feel a bit unnatural (very Java-ish). E.g., run `git add --help` or `ls --help` (on linux) to see the difference.

I propose we fix this now, before the cmdline API is frozen for v1.0 -- it will be very hard (impossible) to change later. The fix is easy (I'm volunteering to do it).

<img width="940" alt="image" src="https://github.com/user-attachments/assets/88becc29-08c1-4e8a-b899-59e6f2cff81b">

mjuric:
Current WIP prototype (early comments welcome):
<img width="1264" alt="image" src="https://github.com/user-attachments/assets/6927bd65-a295-45cc-ad74-d0e9904151ab">

mschwamb:
command line argument --help needs to show what parameters are required and which are optional

mschwamb:
@mjuric  since we have a full draft of the paper and we have a list of variable names - can you get to this, since you're the one who wanted this?

mjuric:
By when would you need it (assessing what's possible w. my schedule these days...)?

mschwamb:
end of November - for submission of the paper

mjuric:
See the PR #1064 -- this implements the minimal changes requested (plus I changed `--orbit` to `--orbits`).

Notes:
* codecov is complaining for unknown reasons; I'll leave you to figure it out (I failed :( ).
* Having a `--verbose` option which defaults to `true` and has to be turned off as `--verbose=false` is double-plus-super-wrong, IMHO. Super confusing. Can't think of any other piece of code that does it. I'm getting a rash just thinking about how wrong this is. Also, `--verbose` is generally understood to increase console verbosity -- for log file control, something like `--log-level` would feel more natural.
[...]

PR #980, opened 2024-07-12
WIP: Command line interface update

mjuric:
This PR reworks the command line arguments to follow GNU conventions (resolving #978), and updates how we do logging.

Changes:
* Use `-` in long arg names, single characters for short args.
* Change the name of some parameters to (hopefully) better reflect what they're about
* Set reasonable defaults for the outputs, so they don't have to be specified. In particular, if the user doesn't override it with --output-dir or --prefix, the outputs (and the log) will all be stored into `sorcha-results*` files (where the suffixes depend on the format, the file being stored, etc). You can see that in the screenshot (note the `sorcha-results.csv` , s`orcha-results-stats.csv`, and `sorcha-results.log` files).
* Change how we handle the logs -- by default, write a /single/ log file named `sorcha-results.log` and append to it (don't overwrite). This solves the problem of accumulating many inconveniently named log files generated for every run. The user can still easily check if there were errors in the log by grepping for "ERROR".

An example of how it all works is shown in the first attached screenshot.

When it comes to logging, we now log the PID on each line, so it's easy to pull out a specific run with something like `grep 68218 sorcha-results.log`. The second screenshot shows what the log looks like now.

image
image

### DIFF
--- a/src/sorcha/modules/PPGetLogger.py
+++ b/src/sorcha/modules/PPGetLogger.py
@@ -4,32 +4,24 @@ from datetime import datetime
 
 
 def PPGetLogger(
-    log_location,
-    log_format="%(asctime)s %(name)-12s %(levelname)-8s %(message)s ",
+    log_fn,
+    log_format="[%(asctime)s|%(process)d] %(levelname)s [%(name)s:%(lineno)d] %(message)s",
     log_name="",
-    log_file_info="sorcha.log",
-    log_file_error="sorcha.err",
 ):
     """
     Initialises log and error files.
 
     Parameters
     -----------
-    log_location : string
-        Filepath to directory in which to save logs.
+    log_fn : string
+        Path to file into which to save the logs.
 
     log_format : string, optional
         Format for log filename.
-        Default = "%(asctime)s %(name)-12s %(levelname)-8s %(message)s "
+        Default = "[%(asctime)s|%(process)d] %(levelname)s [%(name)s:%(lineno)d] %(message)s",
 
     log_name : string, optional
         Name of log. Default = ""
-
-    log_file_info : string, optional
-        Name with which to save info log. Default = "sorcha.log"
-
-    log_file_error : string, optional
-        Name with which to save error log. Default = "sorcha.err"
 
     Returns
     ----------
@@ -46,21 +38,9 @@ def PPGetLogger(
     # stream_handler.setFormatter(log_formatter)
     # log.addHandler(stream_handler)
 
-    dstr = datetime.now().strftime("%Y-%m-%d-%H-%M-%S")
-    cpid = os.getpid()
-
-    log_file_info = os.path.join(log_location, dstr + "-p" + str(cpid) + "-" + log_file_info)
-    log_file_error = os.path.join(log_location, dstr + "-p" + str(cpid) + "-" + log_file_error)
-
-    file_handler_info = logging.FileHandler(log_file_info, mode="w")
+    file_handler_info = logging.FileHandler(log_fn, mode="a")
     file_handler_info.setFormatter(log_formatter)
-    file_handler_info.setLevel(logging.INFO)
     log.addHandler(file_handler_info)
-
-    file_handler_error = logging.FileHandler(log_file_error, mode="w")
-    file_handler_error.setFormatter(log_formatter)
-    file_handler_error.setLevel(logging.ERROR)
-    log.addHandler(file_handler_error)
 
     log.setLevel(logging.INFO)
 
--- a/src/sorcha/modules/PPCommandLineParser.py
+++ b/src/sorcha/modules/PPCommandLineParser.py
@@ -5,7 +5,7 @@ import glob
 from .PPConfigParser import PPFindFileOrExit, PPFindDirectoryOrExit
 
 
-def warn_or_remove_file(filepath, force_remove, pplogger):
+def warn_or_remove_file(filepath, force_remove, pplogger, exclude_remove=[]):
     """Given a path to a file(s), first determine if the file exists. If it does not
     exist, pass through.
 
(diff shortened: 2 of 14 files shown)


mschwamb:
I am all for the changes in principle. I have some concerns I would like to see addressed. 

Having to write out the sorcha command a few times for my runs manually, this is now much more I have to write. Like --integrator-data be -i ? and --force be -f ? 

I find the categories input, output, and advanced confusing - it's not what you get with most man pages/--help so I am happy to have changes made but then I strongly prefer required and optional be the two categories, rather than  inputs, outputs, advanced. It's not what most people are used to when they look at man pages and --help for commands. Also -f forcing output to overwrite existing files is not really an advanced feature so I find it confusing it is listed in advanced. In the current version of the code base there is the ability to make the tally file (now --stats in this PR). It is an optional output. It is not required.  From the way the current arguments are listed I would think it is required.  So can we please use required and optional arguments for the classes here. 

We call one of the input files he physical parameters  (colors, H, phase curves) file and the complex parameters file (things that add-ons needs like for light curve or cometary activity parameters). I really do not want to have --colors because that's going to be confusing when the rest of the help document and the paper talks about the physical parameters file  and that file contains crucially H, so I would prefer it not be called the colors file. 

same thing for --extra-object-data  - I do not intuitively know what that means and give it's called complex physical parameters in the documentation I would prefer we don't change it here. 

I don't love --intergrator-data though maybe it's  - maybe it's better than what we currently have.

mschwamb:
A strong no to the change in the log. I found it helpful in my runs to have separate log files per run. I somehow had the super computer running processes twice and it was causing issues with my output and it was easy to figure out what happened because two log files appeared in my run directory - I believe it this case they would combine into one log file and I would have to figure out by looking there are two processes IDs

The multiple log files is only an issue for Rubin operations. I think you can find away around this without changing the code as its primarily job is for running on HPCs and mistakes like running processes twice can happen. seeing multiple log files makes it easy to see what happened and also someone may push all output for simulations to the ame output directory which would then cause havoc on the log file.

src/sorcha/modules/PPGetLogger.py
```diff
@@ -4,33 +4,25 @@
 
 
 def PPGetLogger(
-    log_location,
-    log_format="%(asctime)s %(name)-12s %(levelname)-8s %(message)s ",
+    log_fn,
+    log_format="[%(asctime)s|%(process)d] %(levelname)s [%(name)s:%(lineno)d] %(message)s",
```
mschwamb, line 8: process id needs to be in the log file name so I am against adding it here

mjuric:
> Having to write out the sorcha command a few times for my runs manually, this is now much more I have to write. Like --integrator-data be -i ? and --force be -f ?
> 

Easy to add `-f` (I'll do it), but think about it -- if you keep using `--force` to the point where you'd like to shorten it, shouldn't clobbering be the default?

> I find the categories input, output, and advanced confusing - it's not what you get with most man pages/--help so I am happy to have changes made but then I strongly prefer required and optional be the two categories, rather than inputs, outputs, advanced. It's not what most people are used to when they look at man pages and --help for commands. Also -f forcing output to overwrite existing files is not really an advanced feature so I find it confusing it is listed in advanced. In the current version of the code base there is the ability to make the tally file (now --stats in this PR). It is an optional output. It is not required. From the way the current arguments are listed I would think it is required. So can we please use required and optional arguments for the classes here.

Sounds good -- that's just help output grouping, easy to change.

> We call one of the input files he physical parameters (colors, H, phase curves) file and the complex parameters file (things that add-ons needs like for light curve or cometary activity parameters). I really do not want to have --colors because that's going to be confusing when the rest of the help document and the paper talks about the physical parameters file and that file contains crucially H, so I would prefer it not be called the colors file.
> 

How about `--photometric-parameters` (or just `--photometry`, or `--phot-params`, or even `--seds`... the shorter the better, IMHO)?

To a dynamicist physical parameters are the mass, density, diameter, shape (or moments), strength, etc. (e.g. googling "asteroid physical parameters" surfaces [this](https://en.wikipedia.org/wiki/Standard_asteroid_physical_characteristics) and [this](https://www.britannica.com/science/asteroid/Physical-characteristics-of-asteroids) as first hits).

> same thing for --extra-object-data - I do not intuitively know what that means and give it's called complex physical parameters in the documentation I would prefer we don't change it here.
> 

How about `--complex-photometric-parameters` (or `--complex-phot-params`)?

Good point about the docs -- I forgot to update them.

mjuric:
For the logs, it comes from me getting disoriented with the number of files generated in a sorcha dir, as every invocation generates a new log file (including when you get the command-line arguments wrong). It's a pain to keep cleaning those up... willing to bet some small amount the end-users will hate it too :).

That said, you're totally right we need to have jobs write into separate log files (and outputs!) for each invocation. This is what `--prefix` is meant to take care of in the above scheme. Specifying `--prefix` changes the prefix of all outputs (what you called the "stem"), and you can use this in scheduler scripts. E.g., with SLURM, you'd write something like:
```
sorcha run --prefix="sorcha-outputs-$SLURM_JOB_ID" ...
```
in the submission script, which would then generate (for example, for job ID 42):
```
sorcha-outputs-42.csv
sorcha-outputs-42-stats.csv
sorcha-outputs-42.log
```
(allowing all jobs to share the same directory for output & logging).

Some options:
1. We could retain the previous behavior, but add a `--log-file [dest.log]` switch which if present would redirect the log to (e.g) `dest.log`. My thinking was that most folks want something like this anyway, so making it the default seemed to make sense. Users can then use `--prefix` to cleanly manage cluster runs (not just for logs but for outputs as well).
1. Or we could default to the new behavior (store the log into `<prefix>.log` by default), but add a command-line argument to activate the previous behavior if that's preferable for one's use case. E.g., something like `--log-dest PID` (or some better name?).
1. Something else...

mschwamb:
`I think I like --phot-params and --complex-phot-params`

We've had lots of discussions on the names of these files last July. Physical parameters and complex physical parameters is the names we came up with. We're not going to change them now.  There's lots in the documentation and the paper to finish - that the priority needs to be on finishing this rather than should we rename files.

mschwamb:
So let's keep physical parameters and complex physical parameters as the descriptors for the non-orbit input files for the simulated populations

mschwamb:
`That said, you're totally right we need to have jobs write into separate log files (and outputs!) for each invocation. This is what --prefix is meant to take care of in the above scheme. Specifying --prefix changes the prefix of all outputs (what you called the "stem"), and you can use this in scheduler scripts. E.g., with SLURM, you'd write something like...`

I somehow accidentally ran multiple identical sorcha commands on bura (strange things are going on when looking at what jobs are running on bura) and because I was using csv they just wrote on top of eachother/appended together. The only reason I new something was up was because I had two log files. Your method would means the log would append and I would have to actually read and understand the format of the output to understand what was happening. Seeing two log files actively being updated in the directory told me exactly what was going on.  I think I had -f on, but I imagine most people want to have the -f option so we shouldn't get rid of it but the log files give the safety of a double check.  Your proposed method would mean I would have to take a detailed look at the log file. Most people will not. 

I can maybe see the argument that the log will should have the prefix but the process ID needs to be there so it's always unique. 

Happy to discuss on a call rather than here.

mjuric:
Allright, sounds like there's a no-go on this... 😢

mschwamb:
Closing again because aspects of this will be taken

PR #990, opened 2024-07-30
Extra checks for empty dataframes

### LINKED ISSUE #986: Check that zero-output results are consistent
Zero-row outputs are generated if bright detections are made but don't survive the linking filter, but don't get made if no bright detections are made. Now that I think of it, doesn't the latest version of Sorcha have the linking filter add a column rather than filter detections? Not sure if this will be reproduced on the latest version of Sorcha, sorry... Anyway, the discrepancy I saw follows.

`2024-07-28 01:20:34,832 sorcha.sorcha INFO     Number of rows BEFORE applying bright limit filter 0 `
`2024-07-28 01:20:34,833 sorcha.sorcha INFO     Number of rows AFTER applying bright limit filter 0 `
`2024-07-28 01:20:34,833 sorcha.sorcha INFO     No observations left in chunk. Skipping to next chunk... `
`2024-07-28 01:20:34,833 sorcha.sorcha INFO     Sorcha process is completed.`
versus
`2024-07-28 01:21:13,175 sorcha.sorcha INFO     Number of rows BEFORE applying bright limit filter 1 `
`2024-07-28 01:21:13,176 sorcha.sorcha INFO     Number of rows AFTER applying bright limit filter 1 `
`2024-07-28 01:21:13,176 sorcha.sorcha INFO     Applying SSP linking filter... `
`2024-07-28 01:21:13,176 sorcha.sorcha INFO     Number of rows BEFORE applying SSP linking filter: 1 `
`2024-07-28 01:21:13,472 sorcha.sorcha INFO     Number of rows AFTER applying SSP linking filter: 0 `
`2024-07-28 01:21:13,472 sorcha.sorcha INFO     Post processing completed for this chunk `
`2024-07-28 01:21:13,472 sorcha.sorcha INFO     Output results for this chunk `
`2024-07-28 01:21:13,473 sorcha.modules.PPOutput INFO     Constructing output path... `
`2024-07-28 01:21:13,473 sorcha.modules.PPOutput INFO     Output to CSV file... `
`2024-07-28 01:21:13,479 sorcha.sorcha INFO     Sorcha process is completed. `

astronomerritt:
Fixes #986.
- Behaviour in sorcha.py regarding empty dataframes is now consistent.
- Each filter checks to make sure the dataframe is not empty before running.
- The code also no longer tries to write output if the dataframe is empty.
- Moved check for `configs["camera_model"] == "none"` out of PPApplyFOVFilter and into sorcha.py, for consistency
- Moved log messages for number of lines left in dataframe away from PPAddUncertainties (which does not remove rows) to PPRandomizeAstrometryAndPhotometry (which does).
- Made sure the loop counter updates in the case where there are no observations in the ephemeris file for the chunk so the chunk is skipped.
- Separated the indexing of SQL output into its own function which is always performed on the last chunk.

### DIFF
--- a/src/sorcha/modules/PPOutput.py
+++ b/src/sorcha/modules/PPOutput.py
@@ -67,7 +67,7 @@ def PPOutWriteHDF5(pp_results, outf, keyin):
     return of
 
 
-def PPOutWriteSqlite3(pp_results, outf, lastchunk=False, tablename="sorcha_results"):
+def PPOutWriteSqlite3(pp_results, outf, tablename="sorcha_results"):
     """
     Writes a pandas dataframe out to a CSV file at a location given by the user.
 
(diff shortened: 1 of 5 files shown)


src/sorcha/sorcha.py
```diff
@@ -332,16 +324,19 @@ def runLSSTSimulation(args, configs):
...
+            PPWriteOutput(args, configs, observations, endChunk, verbose=args.verbose)
+            if args.stats is not None:
+                stats(observations, args.stats, args.outpath, linking=configs["SSP_linking_on"])
+        else:
+            verboselog("No observations left in chunk. No output will be written for this chunk.")
 
-        if args.stats is not None:
-            stats(observations, args.stats, args.outpath, linking=configs["SSP_linking_on"])
+        if endChunk >= lenf and configs["output_format"] == "sqlite3":
+            pplogger.info("Last chunk detected. Indexing output SQLite database...")
```
mschwamb, line 338: Why not pull this outside of the loop rather than check if on the last chunk. If the loop completed you finished the last  chunk and can index. Saves a comparison.

src/sorcha/modules/PPOutput.py
```diff
@@ -92,20 +95,37 @@ def PPOutWriteSqlite3(pp_results, outf, lastchunk=False, tablename="sorcha_resul
...
+    tablename: string
+        String of the table within the database to be indexed.
+
+    Returns
+    -----------
+    None.
+
+    """
+
+    cnx = sqlite3.connect(outf)
```
mschwamb, line 119: What if all the chunks are empty and it never ever got here. I'd add a quick check that the file actually exists or just gracefully return out.

src/sorcha/sorcha.py
mschwamb: I think you need to check the file exists because if it doesn't because all the chunks were empty there is no file. I'm not sure what's the best place for that.

mschwamb (approved):
Looks good to me. Given the all the tests are passing and the logic looks good to me. I think can be merged and @Gerenjie can check later.

PR #992, opened 2024-08-05
Adding linking config variables

### LINKED ISSUE #953: Config file variable for dropping unlinked objects
Whether unlinked objects are dropped or not should be controlled via a config file variable - so we either keep everything, or only discovered objects.

### LINKED ISSUE #954: Linked column should be in default output if it exists
Note that if the linking filter is off, the Linked column won't exist in the dataframe.

### LINKED ISSUE #975: Don't hardcode local noon for linking purposes
Emulation of tracklet linking has a concept of nights. A night is defined to include all observation from the current MJD's local-noon, to the next local noon. At present, the local noon is hardcoded to Chile (*). The user should be able to customize it.

Two options:
* easy to implement: just read it from the config file and pass it on `night_start_utc_days` arg to `linkConfig` in `PPLinkingFilter`.
* easy for the end-user: figure out the local noon from the observatory code.

N.b., current mini-difi defaults can be found [here](https://github.com/dirac-institute/sorcha/blob/45b5a4e102f5b7f4a2a6a771b19baf37695be69e/src/sorcha/modules/PPMiniDifi.py#L294)

(*) This is not quite true...  we currently hardcode 2pm in Chile as the night boundary, but it may make sense to shift this to noon.

_Originally posted by @mjuric in https://github.com/dirac-institute/sorcha/issues/959#issuecomment-2185038858_

astronomerritt:
Fixes #953.
- There is now a "hidden" config file variable, drop_unlinked, that controls whether unlinked objects are dropped by the linking filter or if they remain in the dataframe.
- By default, `drop_unlinked=True`.
- If linking is on and `drop_unlinked=False` then the tally file will contain the linking information. Otherwise it will not.
- Wrote unit tests to ensure expected behaviour in `PPLinkingFilter()` and `PPStats()`.
- Updated unit tests.

Fixes #954.
- If `drop_unlinked=False`, `SSP_linking_on=True` and `output_format=basic`, the object_linked column now appears in the default basic output.
- Wrote a unit test to ensure the column is correctly included.

Fixes #975.
- A new config file variable, SSP_night_start_utc, controls the time in UTC at which the night "starts" at the observatory, which should be local noon.
- This config file variable is mandatory if linking is on.
- Changed all example/demo config files to include this variable with a default of 16.0.
- Updated unit tests.

**NOTE:** I used Chile Standard Time for Chilean noon, hence 4pm UTC. Mario was using Chile Summer Time, which is why he has 2pm Chile = 5pm UTC. I don't know which one is correct here so I defaulted to Chile Standard Time, will change it back if needed.


mschwamb:
@astronomerritt  can you resolve conflicts?

astronomerritt:
Fixed!

mschwamb (approved):

PR #999, opened 2024-08-14
Updating docs regarding disordered output

astronomerritt:
Quick fix to documentation to add a note that when using a pre-generated ephemeris file, Sorcha's output will be in a different order than the output produced when generating the ephemeris within Sorcha, but the output IS otherwise identical.

### DIFF
--- a/docs/ephemerisgen.rst
+++ b/docs/ephemerisgen.rst
@@ -115,7 +115,11 @@ If you want to use the same input orbits across multiple Sorcha runs, you can sa
 
 .. attention::
    Currently the Sorcha-generated ephemeris is outputted in CSV, whitespace or HDF5 file format only.
+   
 
+.. attention::
+   Users should note that output produced by reading in a previously-generated ephemeris file will be in a different order than the output produced when running the ephemeris generator within Sorcha.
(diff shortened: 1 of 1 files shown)


docs/ephemerisgen.rst
```diff
@@ -115,7 +115,11 @@ If you want to use the same input orbits across multiple Sorcha runs, you can sa
 
 .. attention::
    Currently the Sorcha-generated ephemeris is outputted in CSV, whitespace or HDF5 file format only.
+   
 
+.. attention::
```
mschwamb, line 120: This needs to go out the output page. Not here. This is about inputting ones own ephemeris, so it seems confusing for this go here. I think it goes in the output where we say that the output is formatted in X order if an ephemeris file is provided. Can you move this

mschwamb (changes requested):

astronomerritt:
Done. I also added a bit to the outputs section about the statistics/tally file while I was in there, and a bit to the linking filter about the option to drop or retain observations of unlinked objects. If you want any of that changed let me know and I'll do it ASAP.

mschwamb (approved):

PR #1012, opened 2024-09-03
Changes to Sorcha logging, utility script

### LINKED ISSUE #1003: Rename log files to share the output filestem and evaluate successful runs.
**Describe the feature you'd like to see in Sorcha**
Currently, Sorcha log files have long names starting with a date, which are not easily identified to their output files. It would be helpful for the log files to share a filestem with their output files, which would allow for easy identification and diagnosis of runtime issues. 

It would also be helpful to have a more explicit/obvious report of the job's failure/success -- I forget what the proposed solution was in the last meeting, but at the moment I'm seeing 'COMPLETED' as the result of slurm jobs even when Sorcha fails, which has tripped me up before (though it was my failure to not investigate the log files properly).

astronomerritt:
Fixes #1003.
- All log files now share an output prefix with the Sorcha output file they are associated with, to aid location of bad runs.
- Wrote a utility script that will recursively search for all Sorcha log files in a given directory/subdirectories and check to see whether the Sorcha runs were successfully completed. Usage given below.
- Unit tests.

Usage for the new utility script is as follows. I put it in the outputs section of the command line utilities, as that made the most sense.

```
>> sorcha outputs check-logs <top level directory filepath> <optional output filepath>
```

The output file is a .csv consisting of the log filename, a True/False flag as to whether the run was successfully completed, and if not, the last line of the log file.

If you don't supply an output filename, it'll just output everything to the terminal. If all runs were successful, **no output file is created**, as I didn't see the point really. A happy message is printed to the terminal instead.

All unit tests pass on my machine.

### DIFF
--- a/src/sorcha/modules/PPGetLogger.py
+++ b/src/sorcha/modules/PPGetLogger.py
@@ -5,6 +5,7 @@ from datetime import datetime
 
 def PPGetLogger(
     log_location,
+    log_stem,
     log_format="%(asctime)s %(name)-12s %(levelname)-8s %(message)s ",
     log_name="",
     log_file_info="sorcha.log",
@@ -18,6 +19,9 @@ def PPGetLogger(
     log_location : string
         Filepath to directory in which to save logs.
 
+    log_stem : string
+        String output stem used to prefix all Sorcha outputs.
+
     log_format : string, optional
         Format for log filename.
         Default = "%(asctime)s %(name)-12s %(levelname)-8s %(message)s "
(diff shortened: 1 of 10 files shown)


Gerenjie:
Works on my side! I ran a few simulations (tens of thousands of objects) on the new branch, saw that my output files were now helpfully named and readable, and used `sorcha outputs check-logs -f .` to confirm that my runs worked!

mschwamb (changes requested):
doesn't this need to be added to [pyproject.toml](https://github.com/dirac-institute/sorcha/blob/main/pyproject.toml)?

astronomerritt:
Nope! sorcha-outputs is already in pyproject.toml, and I added this command there.

mschwamb (approved):

PR #1024, opened 2024-10-02
Fixing config parser for bright limit

### LINKED ISSUE #1022: The known_objects config file breaks PPConfigParser
In the known_objects config.ini file there is no bright_limit setting - this means that in line 446 of PPConfigParser whenever the parser tries to strip the config "bright_limit" string, it is attempting to strip a None type object and failing. Should probably check beforehand if it exists or not. (Steph says only need to change lines 445-459 to include a check if not None type object)

astronomerritt:
Fixes #1022.

Desired behaviour:
- if the bright limit keyword is not in the config file (e.g. if you are using Rubin_known_object_prediction.ini) the bright limit cut should be skipped

Actual behaviour:
- code broke completely because I forgot to wrap some stuff in an if-statement in PPConfigParser

Fix:
-  put the if-statement in

### DIFF
--- a/src/sorcha/modules/PPConfigParser.py
+++ b/src/sorcha/modules/PPConfigParser.py
@@ -442,21 +442,24 @@ def PPConfigFileParser(configfile, survey_name):
         config, "SATURATION", "bright_limit", "none"
     )
 
-    try:
-        bright_list = [float(e.strip()) for e in bright_limits.split(",")]
-    except ValueError:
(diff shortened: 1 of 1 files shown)


astjoephysics (approved):

astronomerritt:
Please ignore CodeCov, PPConfigParser is impossible to unit test properly in its current form. (One day I'll fix it. ONE DAY.)

PR #1025, opened 2024-10-03
Adding method to check IDs

### LINKED ISSUE #1001: Check what happens when the colors aren't specified for an object
Verify what happens if one runs Sorcha with a catalog of objects, not all of which have entries in the physical characteristics file (the one specified with `-p`).

The version presently on conda-forge doesn't fail in this case, but it's unclear how it computes the magnitudes it outputs.

astronomerritt:
Fixes #1001.
- Added a method, `CombinedObjectReader.check_aux_object_ids()`, that uses existing code to check that the ObjIDs in the auxiliary input files are an exact match for each other.
- If the files do not contain exactly the same ObjIDs, an error is thrown and the code stops.
- This check is performed only ONCE per Sorcha run, and takes 1.4s on input files of 10k lines.
- Wrote a unit test.

Meg: this isn't the way we discussed implementing it. I think it's better. Feel free to disagree :)

Dataframes aren't merged until chunking has begun. If we check dataframe size on merge, there's a possibility that the user could get to chunk 999 of 1000 and THEN have the code fail because the ObjIDs don't match. This would be very annoying.

This method checks once, using code Jeremy already wrote, and takes 1.4s on 10k-row input files. I think it's worth it.


bernardinelli (approved):
Looks good to me!

PR #1047, opened 2024-10-24
created a demo notebook to demonstrate how to calculate LSST bandpass colours

astjoephysics:
Created a new demo notebook that takes (and credits!) spectrum of the TNO 2002 PN34 and walks through how a user could convert it into a colour in the LSST bandpass for their models. Requires rubin_sim installation for the user to run, but for the RTD the outputs should already have been ran and have outputs displayed.


mschwamb (changes requested):
Very close to being complete. You'll need to include it in the documents readthedocs page for the notebook to be visible (editing https://github.com/dirac-institute/sorcha/blob/main/docs/notebooks.rst) and also adding in any packages you need installed to requirements.txt (https://github.com/dirac-institute/sorcha/blob/main/docs/requirements.txt)

astjoephysics:
updated requested files!

mschwamb (approved):

PR #1048, opened 2024-10-24
cleaned up minidifi, added docstrings, updated PPLinkingFilter unit test to adapt to updated minidifi

### LINKED ISSUE #1046: Clean up minidifi  and add doc strings
Functions in mindifi need docstrings and there are some commented out code/main code we want to clean up/remove for consistency of the codebase

astjoephysics:
Fixes #1046

miniDifi has also had remaining bug testing code swept up and removed (🧹 sorry Mario).

In doing so, I removed a segment that updated the default linking config variables locally within PPMiniDifi as per a call on 23rd Oct. This is not needed here as it is read in from the configs variable within PPLinkingFilter -> PPMiniDifi anyways and all results remain the same. This does however affect two unit tests within test_PPLinkingFilter that depended on this config being read in locally. It can and has been fixed by adding the night_start_utc_days as extra arguments to test_PPLinkingFilter_nlink1 and test_PPLinkingFilter_discoveryChances and the unit tests work again.

Also doc strings have now been added to all functions within miniDifi. May want to double check I have described everything accurately just in case.

### DIFF
--- a/benchmarks/bench_cProfile.py
+++ b/benchmarks/bench_cProfile.py
@@ -21,7 +21,7 @@ if __name__ == "__main__":  # pragma: no cover
     # get path to Sorcha top-level folder
     path_to_file = os.path.abspath(__file__)
     path_to_sorcha = str(Path(path_to_file).parents[1])
-    
+
     print(path_to_sorcha)
 
     cmd_args_dict = {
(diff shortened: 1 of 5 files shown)


mschwamb:
@mjuric can you give this a quick scan since it's your handiwork being cleaned up for the JOSS reviewers.  If not @bernardinelli could you take a look?

mjuric:
This looks good to me, OK to merge!

mschwamb (approved):
Because @mjuric  said so

PR #1063, opened 2024-11-19
Clarifying error message in PPReadPointingDatabase()

### LINKED ISSUE #1058: PPReadPointingDatabase shouldn't except all errors
We have this in `PPReadPointingDatabase()`:

https://github.com/dirac-institute/sorcha/blob/7bcb8b6fa320436d79d2c8b97f7928185c37baf8/src/sorcha/modules/PPReadPointingDatabase.py#L33-L41

`except Exception` is bad practice. That except clause is triggering no matter what goes wrong with the pointing database read.

It should be replaced with `except pd.errors.DatabaseError`. According to [Pandas docs](https://pandas.pydata.org/docs/dev/reference/api/pandas.errors.DatabaseError.html) this error is thrown only when the SQL query is bad. Other errors will not be mistakenly caught by this clause.

astronomerritt:
Fixes #1058.
- Split up exception handling in `PPReadPointingDatabase()`.
- One error message triggers specifically if the SQL query is problematic. This would be the most common cause of failure.
- However, Joe has reported a failure for a different, unknown reason (probably file access issues due to multiple HPC runs). The second except catches all other exceptions and will include the Python error message in the message it prints to the terminal and log.

Describe your changes.

### DIFF
--- a/src/sorcha/modules/PPReadPointingDatabase.py
+++ b/src/sorcha/modules/PPReadPointingDatabase.py
@@ -32,13 +32,16 @@ def PPReadPointingDatabase(bsdbname, observing_filters, dbquery, surveyname):
 
     try:
         df = pd.read_sql_query(dbquery, con)
-    except Exception:
+    except pd.errors.DatabaseError:
         pplogger.error(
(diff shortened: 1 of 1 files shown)


mschwamb (approved):
LGTM

PR #1072, opened 2024-11-25
537 config class

### LINKED ISSUE #537: Change the config file parser to a data class
Create a dataclass to replace the current config file parser. The current code there works, but it could be optimized/made easier to add various config file variables.

### LINKED ISSUE #1075: config file reader allows you to have all the stuff for ephemeris generation specified but set ephemeris generation to external
config file reader allows you to have all the stuff for ephemeris generation specified but set ephemeris generation to external

Little-Ryugu:
Fixes #537 
Fixes #1075 

make a config class.


mschwamb:
@astronomerritt  I’m not sure [these function](https://github.com/dirac-institute/sorcha/blob/017198fe157fe7a8c1daefbc4fc2c2f24a161755/src/sorcha/utilities/sorchaConfigs.py#L79) calls in the if statement  are actually doing anything given the floats and other variables are being initialized with default values at the start of each class . Can you double check?

tests/ephemeris/test_ephemeris_generation.py
```diff
@@ -123,14 +123,14 @@ def test_ephemeris_end2end(single_synthetic_pointing, tmp_path):
     pplogger = PPGetLogger(cmd_args_dict["outpath"], "test_log")
     args = sorchaArguments(cmd_args_dict)
 
-    configs = PPConfigFileParser(
+    configs = sorchaConfigs(
         args.configfile,
         args.surveyname,
     )
-    configs["seed"] = 24601
+    #configs["seed"] = 24601
```
mschwamb, line 130: if this is not being used get rid of the line don’t comment it out

tests/data/PPConfig_goldens_test.ini
mschwamb, line 1: these config files should not be changing. What happened?

Little-Ryugu: I changed the config files because I had checks in the fovConfig making footprint_edge_threshold mandatory and also for when ephemerides_type is ar in simulationConfigs. I can remove those checks and then they should be able to pass those unit tests without changing the golden config files.

mschwamb:
So looking at where pytest is failing. Nicely done spotting two test config files were we specified it was running with an external ephemeris file but we had the parameters for A+R ephemeris generation uncommented out. So I've commented those out those lines, because these tests are using a pre-computed ephemeris file.  I've pushed that change back to the branch. 

So now all the failed tests are dying with SystemExit: ERROR: No value found for required key footprint_edge_threshold in config file. Please check the file and try again.

footprint_edge_threshold is not required every time the footprint filter is run. If it is specified in the config then it triggers the more complicated footprint filter. If it's not present then that's fine and the footprint filter is run that just deals with the detector and goes to the edge of the detector, not the additional bit to cut around the edge.  So this should be adjusted in the config file data class.  If footprint_edge_threshold is specified then you need all the parameters of rest of the more detailed footprint camera filter, but you can use the camera fooprint filter without having this config parameter included.

tests/data/PPConfig_goldens_test.ini
mschwamb: checking ephemerides_type is ar when the other bits about the ephemeris simulation are specified is correct - you spotted a edge case the old config file parser missed.

mschwamb: footprint_edge_threshold is not mandatory. It's set to None if the old config parser if it does not exist in the config file and it being none triggers the simpler camera footprint that deals with detectors.  So we'll need that change made

PS (not need to respond to me outside of hours - evenings or weekends may be the only time I can take a look)

astronomerritt:
> @astronomerritt I’m not sure [these function](https://github.com/dirac-institute/sorcha/blob/017198fe157fe7a8c1daefbc4fc2c2f24a161755/src/sorcha/utilities/sorchaConfigs.py#L79) calls in the if statement are actually doing anything given the floats and other variables are being initialized with default values at the start of each class . Can you double check?

When the simulationConfigs object is created in sorchaConfigs, it's being [initialised with a dictionary of actual values](https://github.com/dirac-institute/sorcha/blob/017198fe157fe7a8c1daefbc4fc2c2f24a161755/src/sorcha/utilities/sorchaConfigs.py#L700), so the if statement calls do trigger. If you were to initialise an empty instance of simulationConfigs then yeah, it wouldn't trigger, but nobody should ever be doing that outside of (potentially) testing.

If you don't like this behaviour then the `self._validate_output_configs()` method could be moved out of `__post_init__` and called explicitly instead of automatically after initialisation.

astronomerritt:
This looks great though, good work.

The methods on the classes need to have properly formatted docstrings, however.

mschwamb:
@astronomerritt can you modify one of the methods to show appropriate doc string format?

astronomerritt:
> @astronomerritt can you modify one of the methods to show appropriate doc string format?

Ryan already Slacked me to ask, I think he's got it :)

post_readin_ephem_nonprimary.csv
astronomerritt, line 1: This file shouldn't be in the repo - looks like someone (me) left a test statement in sorcha.py which creates it.

src/sorcha/utilities/sorchaConfigs.py
```diff
@@ -0,0 +1,1272 @@
...
+import numpy as np
+from sorcha.lightcurves.lightcurve_registration import LC_METHODS
+from sorcha.activity.activity_registration import CA_METHODS
+
+
+@dataclass
+class inputConfigs:
+    """Data class for holding INPUTS section configuration file keys and validating them."""
+
+    ephemerides_type: str = ""
```
astronomerritt, line 15: Meg has convinced me that all defaults should be None, so:

`ephemerides_type: str = None`

This makes it clearer that the attribute is unpopulated. Empty strings and zeros do evaluate as False in Python, but they can also cause confusion - especially when a zero (or an empty string) is a legitimate, wanted value.

src/sorcha/utilities/sorchaConfigs.py
```diff
@@ -0,0 +1,1272 @@
...
+    reason : string
+        reason given in the error message on why this value shouldn't be in the config file
+
+    Returns
+    ----------
+    None.
+    """
+
+    # checks to make sure value doesn't exist
+    if value:
```
astronomerritt, line 941: Likewise this should be:

`if value is not None:`

Basically we shouldn't be relying on Python evaluating 0 and "" as False, we should explicitly use None for unpopulated attributes.

mschwamb:
I've deleted the temp csv file in the PR that was being created

src/sorcha/utilities/sorchaConfigs.py
```diff
@@ -0,0 +1,1272 @@
...
+
+        Parameters
+        -----------
+        None.
+
+        Returns
+        ----------
+        None
+        """
+        # make sure all the mandatory keys have been populated.
```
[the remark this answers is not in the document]
Little-Ryugu: self._ephmerides_type is read into simultionConfigs from the inputConfig class, so a check has already been made to ensure that it is in that list. Should I create the check anyway for when simulationConfigs is tested independently? 

I put an underscore in front of it to indicate that it should only be used for checks in simulationConfigs because ephemerides_type is called from the inputConfigs class for the rest of the code.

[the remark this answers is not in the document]
astronomerritt: Oh, I see what you've done! Nah, you're good, that's good practice.

astronomerritt:
I'm very happy with this. I've approved. @mschwamb I will leave it to you to merge in case there's anything you want to change.

astronomerritt (approved):

PR #1084, opened 2024-12-30
Correct the spelling of Steve Chesley's name

ijiraq:
The name 'Chesley' appears in the comments but it should be 'Chesley' 

No code changes, just comments.


mschwamb:
Thanks for catching that @ijiraq

mschwamb (approved):

PR #1087, opened 2025-01-08
Minor clean up issues related to including the config file data classes

### LINKED ISSUE #1081: minor clean up issues related to including the config file data classes
- [ ] sorchaConfigs.py has a whole bunch of functions that replace using other functions like PPGetIntOrExit() confirm if we still need PPGetIntOrExit() if it such functions should be removed
- [ ] remove commented out line https://github.com/dirac-institute/sorcha/blob/6ff172be90daa2160ce7eb0bf9b64765926bd2e7/src/sorcha_cmdline/run.py#L166
- [ ] remove unused functions in [PPConfigParser.py](https://github.com/dirac-institute/sorcha/
blob/6ff172be90daa2160ce7eb0bf9b64765926bd2e7/src/sorcha/modules/PPConfigParser.py#L348) that are not used
- [ ] remove https://github.com/dirac-institute/sorcha/blob/6ff172be90daa2160ce7eb0bf9b64765926bd2e7/src/sorcha_cmdline/run.py#L15
- [ ] ensure that all config variables in the data classes are outputted to the log 
- [ ] remove all unused old function tests related to the old method for parsing the config file

Little-Ryugu:
Fixes #1081 .

Removed some commented out lines, removed unused functions and their unit tests in PPConfigParser.py, made sure all config variables are outputted to log and a unit test has been made for function PrintConfigsToLog.


src/sorcha/modules/PPConfigParser.py
mschwamb, line 1: I think it's confusing to keep a file that's called PPConfigParser with just these functions in them - either we should rename this so it's a file utility or something but ebing PPConfigParser is probably not informative or helpful in the long term for maintenance.

mschwamb: What about shoving this into utilities folder with a different name?

mschwamb (changes requested):
minor change - I think it's PPConfigParser goes away as a file for clarity of what is going on in that function

mschwamb:
Looks good to me.

mschwamb (approved):

PR #1089, opened 2025-01-09
Changing mag limit cut to use trailed source magnitude

### LINKED ISSUE #1083: move mag and SNR cuts to the end after linking?
[Right now the mag and SNR cuts occur before ](https://github.com/dirac-institute/sorcha/blob/6ff172be90daa2160ce7eb0bf9b64765926bd2e7/src/sorcha/sorcha.py#L282C1-L292C106)the fading function so you can't reliably using the linking filter. Especially with the new filter to cut out observations that are far in apparent magnitude before emphemeris generation, we don't really need to make post-processing faster that I think was the original intent. Note this is not the 2 SNR cut. 

Where these cuts are placed this is good for checking if a known orbit will ever get bright enough to be  potentially observed in past LSST observations,  but you might want to simulate a Col-OSSOS sample and you can't do that. We also might want to give options for the mag cut whether it's median, max, min of all observations or the discovery observations.

Little-Ryugu:
Fixes #1083.

Changed mag limit cut to use trailed source magnitude instead of PSF mag and updated unit test.

### DIFF
--- a/tests/sorcha/test_sorchaConfigs.py
+++ b/tests/sorcha/test_sorchaConfigs.py
@@ -963,6 +963,19 @@ def test_activity_config():
 
 # expert config test
 
+@pytest.mark.parametrize("key_name", ["SNR_limit", "mag_limit"])
+def test_expert_config_float(key_name):
+    """
+    tests that wrong inputs for expertConfigs float attributes is caught correctly
+    """
+
+    expect_configs = correct_expert.copy()
+    expect_configs[key_name] = "str"
+    with pytest.raises(SystemExit) as error_text:
+        test_configs = expertConfigs(**expect_configs)
+
+    assert error_text.value.code == f"ERROR: expected a float for config parameter {key_name}. Check value in config file."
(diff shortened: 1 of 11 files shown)


mschwamb (approved):

tests/sorcha/test_sorchaConfigs.py
```diff
@@ -883,7 +883,7 @@ def test_outputConfigs_inlist(key_name, expected_list):
 @pytest.mark.parametrize("key_name", ["position_decimals", "magnitude_decimals"])
 def test_outputConfigs_decimel_check(key_name):
     """
-    Checks that if deciamels are not float or are negative it is caught correctly
+    Checks that if deciamels are not int or are negative it is caught correctly
```
mschwamb, line 886: deciamels --> decimals

mschwamb (changes requested):
minor typo to fix

mschwamb (approved):

Issue #1067, opened 2024-11-22
Should we add a filter that auto-culls input observations that will never be observable?


astjoephysics:
We can calculate the Hmag at the LSST limiting magnitude per object - do we then need to expend compute resources to simulate them? Maybe we can add another column that flags whether or not the object was simulated or not in that case?

PR #1029, opened 2024-10-09
Add a new unit test to make sure the correct linking window has been applied within miniDifi

astjoephysics:
Fixes #1009 #1010.

PR #1094, opened 2025-01-13
Added a culling filter for faint objects

astjoephysics:
Adds a culling filter (#1067) that gets rid of faint objects before ephemeris generation (thank you to Steph+Pedro for developing the majority of this!). Adds a unit test as well for this + additional functions on grabbing perihelion rows and orbit conversion. Also lints a lot of unlinted files.

Also fixes a small bug in unit test for applying linking window (#1029) that wasn't caught - function returned before the rest of the unit test carried out. Didn't fail the unit test but wasn't completely being ran before.


mschwamb:
I don't know what's going on, there's a huge bunch of tests that are being update with spaces. It makes it difficult to review this.

mschwamb:
Maybe it was the linter. Maybe black preferred some changes once run?  - There's one conflict now needing to be to deal with, but it looks pretty straight forward to resolve

mschwamb:
Looks like the conflict is linting

mschwamb (approved):

PR #1152, opened 2025-06-01
Output Docs Review

Gerenjie:
Finally getting around to my output docs job… 
* The columns object_linked and date_linked_MJD are missing from the [] but incorrectly present in the The Basic Output Column Names table lists, so I’ve moved them. 
* fieldID was missing from the ‘all’ and ephemeris output columns, so I’ve added it.
* The Example Detections File in Basic Format appear quite different from what I get in the recent run, so I’ve replaced it with the top ten lines of the demo output.
* The stats file does not appear to include object_linked so I’ve updated the docs to match. It does include date_linked_MJD, though.
* A description of -ew output columns is missing both from this and from the “Ephemeris Generation” page, so I’ve added it here. 


docs/outputs.rst
```diff
@@ -189,19 +189,19 @@ Detections File: Full Output Column Names, Formats, and Descriptions
...
 | phase_deg                          | Float        | The sun-object-observer angle (degrees)                                                                  |
 +------------------------------------+--------------+----------------------------------------------------------------------------------------------------------+
-| epochMJD_TDB                       | Float        | Epoch of orbit (MJD) in Barycentric Dynamical Time                                                       |
+| q                                  | Float        | Object perihelion (au)                                                                                   |
 +------------------------------------+--------------+----------------------------------------------------------------------------------------------------------+
-| t_p_MJD_TDB                        | Float        | Time of periapsis (MJD)  in Barycentric Dynamical Time                                                   |
+| e                                  | Float        | Orbital eccentricity                                                                                     |
```
mschwamb, line 194: this depends on what you're doing - I don't have a q because I inputted Keplerian orbit - so I think we need to say orbital elements

docs/outputs.rst
```diff
@@ -189,19 +189,19 @@ Detections File: Full Output Column Names, Formats, and Descriptions
...
-| argPeri                            | Float        | Argument of periaspsis of the object (degrees)                                                           |
+| inc                                | Float        | Orbital inclination (degrees)                                                                            |
 +------------------------------------+--------------+----------------------------------------------------------------------------------------------------------+
 | node                               | Float        | Longitude of the ascending node of the object (degrees)                                                  |
 +------------------------------------+--------------+----------------------------------------------------------------------------------------------------------+
-| inc                                | Float        | Orbital inclination (degrees)                                                                            |
+| argPeri                            | Float        | Argument of periaspsis of the object (degrees)                                                           |
```
mschwamb, line 200: this depends on what you're doing - I don't have a q because I inputted Keplerian orbit - so I think we need to say orbital elements

docs/outputs.rst
```diff
@@ -189,19 +189,19 @@ Detections File: Full Output Column Names, Formats, and Descriptions
...
 +------------------------------------+--------------+----------------------------------------------------------------------------------------------------------+
-| e                                  | Float        | Orbital eccentricity                                                                                     |
+| t_p_MJD_TDB                        | Float        | Time of periapsis (MJD)  in Barycentric Dynamical Time                                                   |
```
mschwamb, line 202: this depends on what you're doing - I don't have a q because I inputted Keplerian orbit - so I think we need to say orbital elements

docs/outputs.rst
```diff
@@ -189,19 +189,19 @@ Detections File: Full Output Column Names, Formats, and Descriptions
...
 +------------------------------------+--------------+----------------------------------------------------------------------------------------------------------+
-| q                                  | Float        | Object perihelion (au)                                                                                   |
+| epochMJD_TDB                       | Float        | Epoch of orbit (MJD) in Barycentric Dynamical Time                                                       |
 +------------------------------------+--------------+----------------------------------------------------------------------------------------------------------+
 | FORMAT                             | Float        | Orbit format string (COM for heliocentric, BCOM for barycentric, KEP for Keplerian, CART for Cartesian)  |
 +------------------------------------+--------------+----------------------------------------------------------------------------------------------------------+
```
mschwamb, line 207: Same thing as above we might want to say orbits

docs/outputs.rst
```diff
@@ -255,6 +255,10 @@ Detections File: Full Output Column Names, Formats, and Descriptions
...
 | Obj_Sun_LTC_km                     | Float        | Object-sun light-time-corrected distance (km)                                                            |
 +------------------------------------+--------------+----------------------------------------------------------------------------------------------------------+
+| object_linked                      | Boolean      | True/False whether the object passed the linking filter. See note below                                  |
++------------------------------------+--------------+----------------------------------------------------------------------------------------------------------+
+| date_linked_MJD                    | Boolean      | MJD (TAI) Date the object was linked (if it was linked) See note below                                   |
++------------------------------------+--------------+----------------------------------------------------------------------------------------------------------+
 
 .. note::
    All positions, positions, and velocities are in respect to J2000.
```
mschwamb, line 264: Let's have a note about object_linked and date_linked_MJD

docs/outputs.rst
```diff
@@ -319,14 +321,74 @@ Ephemeris Output
...
++------------------------------------+--------------+----------------------------------------------------------------------------------------------------------+
+| phase_deg                          | Float        | The sun-object-observer angle (degrees)                                                                  |
++------------------------------------+--------------+----------------------------------------------------------------------------------------------------------+
+
+.. note::
+   All positions, positions, and velocities are in respect to J2000.
+
+.. note::
+   All columns in the comple physicalx parameters file will also be included in the full output.
+
```
mschwamb, line 391: We have this under the inputs section  - we should directly link there, but I think better to not duplicate here otherwise we have things to maintain  - check the inputs is correct - I am pretty sure it is.

mschwamb (changes requested):
Thanks fort doing this and spotting typos and missing columns - a few suggested edits/clarifications and a request about the ephemeris output

docs/outputs.rst
Gerenjie: After some tests, I see them both showing up for both values of `object_linked` and have changed the docs to match.

PR #1167, opened 2025-09-02
Add default-matching drop_unlinked config to .ini files

Gerenjie:


mschwamb:
We'll reassess in early-mid 2026

PR #1181, opened 2026-01-28
Run sorcha without forcing reading config file or forced writing results to disk

yoachim:
For cases where one has several pointing histories and several solar system body populations, it would be useful to loop over those within python and call `runLSSTSimulation` directly without using the command line and configuration file(s).

* Update so `sorchaConfigs` is now a subclass of `basesorchaConfigs`, so if one wants to make a config without reading a file it is possible to with `basesorchaConfigs`, and `sorchaConfigs` behaves as before. 
* Add `return_only` kwargs to `runLSSTSimulation` and `PPStats.stats` so users are not forced to write output to disk in case they just want the DataFrames returned
* Minor fixes in some other config classes so they work with both reading config files and manual instatiation.

- [ x] Does pip install still work?
- [ x] Have you written a unit test for any new functions?
- [ x] Do all the units tests run successfully?
- [ x] Does Sorcha run successfully on a test set of input files/databases?
- [ x] Have you used black on the files you have updated to confirm python programming style guide enforcement?

### DIFF
--- a/src/sorcha/modules/PPStats.py
+++ b/src/sorcha/modules/PPStats.py
@@ -1,7 +1,7 @@
 import os
 
 
-def stats(observations, statsfilename, outpath, sconfigs):
+def stats(observations, statsfilename, outpath, sconfigs, return_only=False):
     """
     Write a summary statistics file including whether each object was linked
     or not within miniDifi, their number of observations, min/max phase angles,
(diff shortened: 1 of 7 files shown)


yoachim:
@mschwamb sounds good, no particular rush. My main pitch here is that I know astronomers love their config and csv files, but it's good practice to let python users call your python code from within python. And if code is going to run on a cluster it's important to give users control over disk I/O. If I've got 30 pointing histories and 7 solar system populations, I don't want 200+ jobs simultaneously trying to dump observations to disk (which I'm then just going to have to read back in and analyze anyway).

Little-Ryugu (changes requested):
Looks nearly good to go. Just a small addition to the docstrings to help future development, and you need to run black version 26.1.0 for the linter.

Little-Ryugu (approved):
Looks all good to me.

PR #1196, opened 2026-04-02
added fix for trailing loss calculations

astjoephysics:
Trailing losses may have a code issue with how they are being calculated (not a maths issue!). TL;DR:

1. Wrong trailing loss model for trailedSourceMagSigma
uncertainties() in PPAddUncertainties.py always calls calcTrailingLoss() with a default model=“circularPSF” (equivalent to finding Δm(PSF+detection)), but Eqn 9 of the paper requires for trailed sources that we use model=“trailedSource” (equivalent to finding Δm(PSF))

2. Double counting detection loss in PSFMagSigma
PSFMagTrue is already constructed via trailedMagTrue + Δm(PSF+detection), but uncertainties() adds another Δm(PSF+detection) due to the above mentioned hardcoding of the circularPSF model before computing SNR, making fast-moving objects appear fainted than they should and be incorrectly dropped by PPSNRLimit

The net result is that for fast movers (really like > ~0.5deg/day or so, the error in our old SNR calculation scales very roughly with v^2) we are overestimating trailing losses, which leads to underestimating SNR, which leads to an underreporting of those objects. Luckily the fix is as simple as making sure we specify which of circularPSF or trailedSource we should be using as a model when calculating trailing losses and making sure they're applied to the correct magnitude, which I think is only 4 or 5 lines of change


astjoephysics:
(RTD failure is a new one...something about skyproj?)

mschwamb:
@bernardinelli I see you updated the tests but our fixes are in a separate branch dmagfix

mschwamb:
Closing this as new PR #1207 has additional changes

PR #1203, opened 2026-06-20
Fix convert_to_geocentric for zero-valued parallax constants

matthewholman:
## Summary

`Observatory.convert_to_geocentric` mistakes a parallax constant that is legitimately `0.0` for a missing position, so several real observatory codes are wrongly reported as having no fixed location.

The check gates on **truthiness**:

```python
if (obs_location.get("Longitude", False)
    and obs_location.get("cos", False)
    and obs_location.get("sin", False)):
```

A constant equal to `0.0` is falsy, so these codes fall through to `(None, None, None)`:

| code | name | zero constant |
|------|------|---------------|
| 500 / 244 / 248 | Geocentric / Occultation / Hipparcos | Longitude = cos = sin = 0 |
| 000 | Greenwich | Longitude = 0 |
| 782 | Quito (equatorial) | sin = 0 |

Downstream consumers then route these to a moving-observatory path that expects per-observation positions and fail on plain MPC input. The common geocentric code **500** is the most impactful.

## Fix

Test for the **presence** of the constants (`is not None`) rather than their truthiness.

- Codes with no parallax keys at all (roving observer, space telescopes) still return `(None, None, None)`.
- The geocenter resolves to a `(0, 0, 0)` offset from Earth's center, which is correct.

## Verification

New test `test_convert_to_geocentric_zero_parallax_constants` covers the geocenter, Greenwich, an equatorial station, and a position-less observatory. It passes with the fix and fails against the previous truthiness check.

Reported downstream in Smithsonian/layup#286.

### DIFF
--- a/src/sorcha/ephemeris/simulation_parsing.py
+++ b/src/sorcha/ephemeris/simulation_parsing.py
@@ -338,10 +338,15 @@ class Observatory:
                 Geocentric position (x,y,z)
         """
         returned_tuple = (None, None, None)
+        # Test for the *presence* of the parallax constants rather than their
+        # truthiness: a constant that is legitimately 0.0 -- the geocenter
+        # (code 500: Longitude=cos=sin=0), Greenwich (code 000: Longitude=0),
+        # or an equatorial station such as Quito (code 782: sin=0) -- is falsy
+        # and would otherwise be mistaken for a missing position.
         if (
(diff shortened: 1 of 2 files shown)


mschwamb:
@bernardinelli  could you take a look at this PR?

bernardinelli (approved):
Looks good to me!

PR #1207, opened 2026-07-01
Trailing loss fixes

### LINKED ISSUE #1206: Bug in our  handling of trailing losses
From @astjoephysics 

Trailing losses may have a code issue with how they are being calculated (not a maths issue!). TL;DR:

    Wrong trailing loss model for trailedSourceMagSigma
    uncertainties() in PPAddUncertainties.py always calls calcTrailingLoss() with a default model=“circularPSF” (equivalent to finding Δm(PSF+detection)), but Eqn 9 of the paper requires for trailed sources that we use model=“trailedSource” (equivalent to finding Δm(PSF))

    Double counting detection loss in PSFMagSigma
    PSFMagTrue is already constructed via trailedMagTrue + Δm(PSF+detection), but uncertainties() adds another Δm(PSF+detection) due to the above mentioned hardcoding of the circularPSF model before computing SNR, making fast-moving objects appear fainted than they should and be incorrectly dropped by PPSNRLimit

The net result is that for fast movers (really like > ~0.5deg/day or so, the error in our old SNR calculation scales very roughly with v^2) we are overestimating trailing losses, which leads to underestimating SNR, which leads to an underreporting of those objects. Luckily the fix is as simple as making sure we specify which of circularPSF or trailedSource we should be using as a model when calculating trailing losses and making sure they're applied to the correct magnitude, which I think is only 4 or 5 lines of change

bernardinelli:
Fixes #1206  .

Describe your changes.


src/sorcha/modules/PPAddUncertainties.py
```diff
@@ -96,91 +97,43 @@ def addUncertainties(detDF, sconfigs, module_rngs, verbose=True):
...
+    detDF["trailedSourceMagTrue"] + dMag, detDF["fiveSigmaDepth_mag"], FWHMeff=detDF["seeingFwhmGeom_arcsec"] * 1000, output_units="mas"
     )
-    photometric_sigma = calcPhotometricUncertainty(SNR)
-    astrSigDeg = (astrSig.values * u.mas).to(u.deg).value
+    detDF["trailedSourceMagSigma"] = calcPhotometricUncertainty(detDF["SNRTrailedSourceMag"])
+    detDF["astrometricSigma_deg"] = (detDF["astrometricSigma_deg"].values * u.mas).to(u.deg).value
 
-    return (astrSigDeg, photometric_sigma, SNR)
+
+    # we don't happy dMag in this case because we're looking at the uncertainty on the PSF mag
```
astjoephysics: typo :^P (happy -> apply)

astjoephysics (approved):

PR #1222, opened 2026-09-12
Fix for 1221

### LINKED ISSUE #1221: Trailing losses getting added to PSFMag in PPDetectionProbability
In the function PPDetectionProbability, trailing losses are being added to the mag. 
https://github.com/dirac-institute/sorcha/blob/4e37cfe974a04a2197c64ff5074dc68608fe1893/src/sorcha/modules/PPDetectionProbability.py#L100-L113

Should this be happening when it's already being added in RunLSSTSimulation at the trailing losses if statement?
https://github.com/dirac-institute/sorcha/blob/4e37cfe974a04a2197c64ff5074dc68608fe1893/src/sorcha/sorcha.py#L254-L264

bernardinelli:
Fixes #1221 .

Describe your changes.


mschwamb (approved):
Looks good to me. I'm surprised this didn't change the goldens. Maybe because this is really small change?

mschwamb:
@Little-Ryugu can you take a look as well starting from the sorcha run main?

Little-Ryugu:
@mschwamb, I was looking there to see why the goldens were the same, and I found this: 

 PPDetectionProbability trailing_losses is false by default.  https://github.com/dirac-institute/sorcha/blob/4e37cfe974a04a2197c64ff5074dc68608fe1893/src/sorcha/modules/PPDetectionProbability.py#L54-L63

and Sorcha calls this function here 
https://github.com/dirac-institute/sorcha/blob/4e37cfe974a04a2197c64ff5074dc68608fe1893/src/sorcha/modules/PPFadingFunctionFilter.py#L38-L43

So it was always false in Sorcha and not affecting anybody's results or goldens. I also checked what happens when this is true by default, and it would just error out due to eph_df["dmagDetect"] not existing.

Little-Ryugu:
Looked in detail from the start of Sorcha there; everything looks good!

mschwamb:
Glad to hear this is more cleanup than a bug fix.  Definitely will make the code less confusing

PR #1224, opened 2026-09-16
Removing trailing loss mentions in PPDetectionProbability

### LINKED ISSUE #1223: Update PPDetectionProbability to remove trailing losses as an optional variable since we don't use it
Currently we removed trailing loss capability because that's already encoded in the PSFMag. So should we remove trailing losses as an optional arguments from PPDetectionProbability?

    trailing_losses : Boolean, optional
        Are trailing losses being applied?, Default = False

    trailing_loss_name : string, optional
        eph_df column name for trailing losses, Default = dmagDetect

Little-Ryugu:
Fixes #1223

Removed function variables and Doc strings not used in  PPDetectionProbability()


src/sorcha/modules/PPDetectionProbability.py
```diff
@@ -53,11 +53,8 @@ def calcDetectionProbability(mag, limmag, fillFactor=1.0, w=0.1):
 
 def PPDetectionProbability(
     eph_df,
-    trailing_losses=False,
-    trailing_loss_name="dmagDetect",
     magnitude_name="PSFMag",
     limiting_magnitude_name="fiveSigmaDepth_mag",
-    field_id_name="FieldID",
```
mschwamb, line 60: Can we keep this one. I can imagine per chip that would be useful?

mschwamb: But my guess is there's additional information needed so maybe that's not most useful on second thought. It could be added in if we're using it again.

mschwamb (approved):

PR #1225, opened 2026-09-16
Order pointing database queries by observation time, not observationId

Gerenjie:
Pointing databases are not necessarily guaranteed to have observationId increasing with observation time. When they don't, ephemeris generation becomes very inefficient.

Order the shipped Rubin/demo/benchmark queries by observationStartMJD_TAI and the DES query by observationMidpointMJD_TAI so the pointings are always read in time order. Docs updated to match.


mschwamb:
Thanks @Gerenjie  One small change before merge. In the troubleshooting page can you add a reference to this. If you're getting really slow results check that you've got your observations from the pointing database returned in order because ephemeris generation will be extremely slow as it is optimized/expecting the observations to be sent to it in sequential order. Or something along those lines. Once that's in. I'll approve and you can merge.

mschwamb (changes requested):
Requesting a small change in the docs so we don't forget about this in the future. Otherwise looks good

